PR #159, opened 2020-04-10
Use ignition-tooling/gazebodistro

scpeters:
This has already been migrated to github,
so use this one instead of bitbucket.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFBase.groovy
@@ -25,7 +25,7 @@ class OSRFBase
           stringParam('RTOOLS_BRANCH','master','release-tool branch to use')
           if (Globals.gazebodistro_branch)
           {
-            stringParam('GAZEBODISTRO_BRANCH','default','gazebodistro branch to use')
+            stringParam('GAZEBODISTRO_BRANCH','master','gazebodistro branch to use')
           }
           booleanParam('NO_MAILS',false,'do not send any notification by mail')
         }
--- a/jenkins-scripts/lib/windows_library.bat
+++ b/jenkins-scripts/lib/windows_library.bat
@@ -141,10 +141,10 @@ goto :EOF
 :: arg2: directory destination (default .)
 set gzdistro_dir=gazebodistro
 
-if "%GAZEBODISTRO_BRANCH%" == "" (set GAZEBODISTRO_BRANCH=default)
+if "%GAZEBODISTRO_BRANCH%" == "" (set GAZEBODISTRO_BRANCH=master)
 
 if exist %gzdistro_dir% (rmdir /s /q %gzdistro_dir%)
-hg clone https://bitbucket.org/osrf/gazebodistro %gzdistro_dir% -b %GAZEBODISTRO_BRANCH%
+git clone https://github.com/ignition-tooling/gazebodistro %gzdistro_dir% -b %GAZEBODISTRO_BRANCH%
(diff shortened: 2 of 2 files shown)


scpeters:
I prefer making GitHub pull requests from a fork, so I did so here, but then I tried to run a test using this branch, and we only support testing from the `origin` repository. I'll push this branch into `origin` for testing

scpeters:
[Build Status](https://build.osrfoundation.org/view/ign-citadel/job/ign_physics-ign-2-win/16/) https://build.osrfoundation.org/view/ign-citadel/job/ign_physics-ign-2-win/16/

j-rivero:
> I prefer making GitHub pull requests from a fork, so I did so here

Using this repo branches makes it a bit faster for me to review changes (not need to clone a new repo just go the usual place and get a new branch).

chapulina:
I was also curious about @scpeters 's preference for forks and wanted to hear more. Sure, git makes it easier to work with several remotes than mercurial does. But I don't see a clear disadvantage to using origin branches rather than forks.

scpeters:
it's just tidier. we have hundreds of open branches on bitbucket because using forks is so much of a pain there, but I find it easier to use forks on GitHub.

If I'm going to test a fork, I usually just add a remote to that local repo (in fact I do this for my own fork):

~~~
git clone https://github.com/ignition-tooling/gazebodistro
cd gazebodistro
git remote add scpeters git@github.com:scpeters/gazebodistro
git fetch scpeters
~~~

It's just one line to add a new remote, so I just add remotes of other people's forks and pull the specific branch I want to test. I use the ssh syntax for the remotes I plan to push to and https for remotes I plan to just read from (for public repos).

~~~
git remote add j-rivero https://github.com/j-rivero/gazebodistro
git fetch j-rivero
git checkout branch_made_by_j-rivero
~~~

j-rivero:
> It's just one line to add a new remote, so I just add remotes of other people's forks and pull the specific branch I want to test.

easy solution thanks, we can go with it, no big deal.

chapulina:
Thanks for the explanation, @scpeters . I can see how it can be tidier, but on the flip side, the code is more scattered around and we always have to remember "where was that branch again?" It also makes it more complicated to collaborate unless we open PRs or share permissions with the entire team.

I think the main reason for the branch multiplication we had on BitBucket was mercurial. And in hindsight, having it all in one place made migration to GitHub "tidier".

I'm not entirely for or against it, I just think there are pros and cons.

scpeters:
I think my original comment was pointing out an issue with my preference for forks, at least with our current CI setup, because you can't just give the branch name to jenkins; you would also have to give the remote name. I agree that there are trade-offs, and I don't think everyone should be forced to use forks.

> the code is more scattered around and we always have to remember "where was that branch again?" It also makes it more complicated to collaborate unless we open PRs or share permissions with the entire team.

Once a PR is open, I think it's not a problem since it is easy to find, and the "Allow maintainers to edit" option handles the permissions. For collaborative branches that aren't ready for a PR (though we could consider [Draft PR's](https://github.blog/2019-02-14-introducing-draft-pull-requests/), which would provide a place for discussion), I think hosting them on the `origin` makes sense. Though many branches don't fall under that category in my opinion. 

> I think the main reason for the branch multiplication we had on BitBucket was mercurial. And in hindsight, having it all in one place made migration to GitHub "tidier".

That's true; we've been ignoring forks for this current migration because of the centralization.

chapulina:
re: many branches: FYI, I enabled auto branch deletion on PR merge for this repo

PR #162, opened 2020-04-10
Use console_cohesion instead of console_direct for colcon logs

j-rivero:
Suggestion by @dirk-thomas to improve the console log.

### DIFF
--- a/jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
+++ b/jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
@@ -24,7 +24,7 @@ else
   export CMD_CATKIN_CONFIG=""
   export CMD_CATKIN_LIST="colcon list -t"
   export CMD_CATKIN_BUILD="colcon build --parallel-workers ${MAKE_JOBS} --symlink-install --event-handler console_cohesion+ ${CATKIN_EXTRA_ARGS}"
-  export CMD_CATKIN_TEST="colcon test --parallel-workers 1 --event-handler console_cohesion+ || true"
+  export CMD_CATKIN_TEST="colcon test --executor sequential --event-handler console_direct+ || true"
   export CMD_CATKIN_TEST_RESULTS="colcon test-result --verbose || true"
   export IGNORE_FILE="COLCON_IGNORE"
 fi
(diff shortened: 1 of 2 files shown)


jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
```diff
@@ -23,8 +23,8 @@ if ${USE_CATKIN_MAKE}; then
 else
   export CMD_CATKIN_CONFIG=""
   export CMD_CATKIN_LIST="colcon list -t"
-  export CMD_CATKIN_BUILD="colcon build --parallel-workers ${MAKE_JOBS} --symlink-install --event-handler console_direct+ ${CATKIN_EXTRA_ARGS}"
-  export CMD_CATKIN_TEST="colcon test --parallel-workers 1 --event-handler console_direct+ || true"
+  export CMD_CATKIN_BUILD="colcon build --parallel-workers ${MAKE_JOBS} --symlink-install --event-handler console_cohesion+ ${CATKIN_EXTRA_ARGS}"
+  export CMD_CATKIN_TEST="colcon test --parallel-workers 1 --event-handler console_cohesion+ || true"
```
dirk-thomas, line 27: In this case where there are no concurrent workers keeping `console_direct+` makes sense to stream the output as it happens.

Same for the `test` invocation in Windows below.

Instead of `--parallel-workers` you could use `--executor sequential` - as it is used below already.

j-rivero: thanks, should be fixed in b44c2b39cd8a6b9a5cc8263c68d11f534314a406

jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
```diff
@@ -23,8 +23,8 @@ if ${USE_CATKIN_MAKE}; then
 else
   export CMD_CATKIN_CONFIG=""
   export CMD_CATKIN_LIST="colcon list -t"
-  export CMD_CATKIN_BUILD="colcon build --parallel-workers ${MAKE_JOBS} --symlink-install --event-handler console_direct+ ${CATKIN_EXTRA_ARGS}"
-  export CMD_CATKIN_TEST="colcon test --parallel-workers 1 --event-handler console_direct+ || true"
+  export CMD_CATKIN_BUILD="colcon build --parallel-workers ${MAKE_JOBS} --symlink-install --event-handler console_cohesion+ ${CATKIN_EXTRA_ARGS}"
+  export CMD_CATKIN_TEST="colcon test --executor sequential --event-handler console_direct+ || true"
```
dirk-thomas, line 27: Actually why are you using `| true` here? I don't think that is desired. The return code only indicates if the tests were involved correctly - not indicating anything about test results.

j-rivero: thanks fixed in a23773f

PR #172, opened 2020-04-17
Port release.py to use github repository as source code

j-rivero:
Port our release tool to be able to work with new github repositories. Code is a bit old and would need a good cleanup but the patch seems to work for me while releasing `ignition-rendering2`.

The use is as usual. Remember to include the `-a ignition-foo2 --ignition-repo` arguments.

Tested:
 * [builds are in Jenkins](https://build.osrfoundation.org/job/ign-rendering2-debbuilder/43)
 * [S3 tarball is in place](https://osrf-distributions.s3.amazonaws.com/ign-rendering/releases/ignition-rendering2-2.4.0.tar.bz2)
 * [tag is ok](https://github.com/ignitionrobotics/ign-rendering/releases/tag/ignition-rendering2_2.4.0)

### DIFF
--- a/release.py
+++ b/release.py
@@ -401,23 +401,31 @@ def generate_upload_tarball(args):
     sourcedir = os.getcwd()
     tmpdir    = tempfile.mkdtemp()
     builddir  = os.path.join(tmpdir, 'build')
+    use_git   = os.path.isdir(os.path.join(sourcedir, '.git'))
+
     # Put the hg-specific stuff in a try block, to allow for a git repo
     try:
         # Check for uncommitted changes; abort if present
-        cmd = ['hg', 'status', '-q']
+
+        cmd = ['git', 'diff-index', 'HEAD'] if use_git else ['hg', 'status', '-q']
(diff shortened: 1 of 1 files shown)


scpeters:
should we just remove the `hg` and `bitbucket` cases at this point?

cc @chapulina

chapulina:
> should we just remove the hg and bitbucket cases at this point?

I'm fine with that. In the worst case scenario, if we need to make a release and the new `git` approach is hiccuping, we can always use an old revision of this repo.

j-rivero:
> should we just remove the `hg` and `bitbucket` cases at this point?

here we go c56d546

scpeters:
I fixed one console message in 912d4f0054f75c7ce0347aa1f65fff86f92a73cb; otherwise it's looking good

I haven't tested it, but we can address any issues when they come up because I think this branch is needed anyway

chapulina:
I just triggered the `ign-physics` 2.1.0 release using this branch and haven't seen anything weird yet.

PR #174, opened 2020-04-20
Don't install Ruby for ign-msgs on Xenial

chapulina:
`ign-msgs1` CI is [failing](https://build.osrfoundation.org/job/ignition_msgs-ci-pr_any-ubuntu_auto-amd64/317/consoleFull) on Xenial with this message:

```
Step 23/34 : RUN gem install protobuf
 ---> Running in 8e959b4281df
ERROR:  Error installing protobuf:
	zeitwerk requires Ruby version >= 2.4.4.
```

A quick solution is to just skip Ruby for Xenial.

Queued a test build: [Build Status](https://build.osrfoundation.org/job/ignition_msgs-ci-pr_any-ubuntu_auto-amd64/325/)

### DIFF
--- a/jenkins-scripts/docker/ign_msgs-compilation.bash
+++ b/jenkins-scripts/docker/ign_msgs-compilation.bash
@@ -17,7 +17,7 @@ fi
 export BUILDING_SOFTWARE_DIRECTORY="ign-msgs"
 export BUILDING_PKG_DEPENDENCIES_VAR_NAME="IGN_MSGS_DEPENDENCIES"
 # skip ruby on Xenial, because protobuf requires Ruby version >= 2.4.4
-export DOCKER_POSTINSTALL_HOOK="gem install protobuf || if [ `lsb_release -sr` == "16.04" ]; then true; else false; fi"
+export DOCKER_POSTINSTALL_HOOK="gem install protobuf || if [ `lsb_release -sr` = "16.04" ]; then true; else false; fi"
 
 # Identify IGN_MSGS_MAJOR_VERSION to help with dependency resolution
 IGN_MSGS_MAJOR_VERSION=$(\
--- a/jenkins-scripts/docker/ign_msgs-compilation.bash
+++ b/jenkins-scripts/docker/ign_msgs-compilation.bash
@@ -16,7 +16,8 @@ fi
 
 export BUILDING_SOFTWARE_DIRECTORY="ign-msgs"
 export BUILDING_PKG_DEPENDENCIES_VAR_NAME="IGN_MSGS_DEPENDENCIES"
-export DOCKER_POSTINSTALL_HOOK="gem install protobuf"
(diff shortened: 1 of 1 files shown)


jenkins-scripts/docker/ign_msgs-compilation.bash
```diff
@@ -16,7 +16,8 @@ fi
 
 export BUILDING_SOFTWARE_DIRECTORY="ign-msgs"
 export BUILDING_PKG_DEPENDENCIES_VAR_NAME="IGN_MSGS_DEPENDENCIES"
-export DOCKER_POSTINSTALL_HOOK="gem install protobuf"
+# skip ruby on Xenial, because protobuf requires Ruby version >= 2.4.4
+export DOCKER_POSTINSTALL_HOOK="gem install protobuf || if [ `lsb_release -sr` == "16.04" ]; then true; else false; fi"
```
j-rivero, line 20:
```suggestion
export DOCKER_POSTINSTALL_HOOK="gem install protobuf || if [ `lsb_release -sr` = "16.04" ]; then true; else false; fi"
```
It is [failing](https://build.osrfoundation.org/job/ignition_msgs-ci-pr_any-ubuntu_auto-amd64/374/console) with 
```
/bin/sh: 1: [: 16.04: unexpected operator
The command '/bin/sh -c gem install protobuf || if [ 16.04 == 16.04 ]; then true; else false; fi' returned a non-zero code: 1
```
/bin/sh does not have a == operator but =. I did not try in Jenkins but should work :)

chapulina: Oh but I tried this locally on a Bionic machine and got this:

```.sh
$ if [ `lsb_release -sr` == "16.04" ]; then true; else false; fi
$ echo $?
1
$ if [ `lsb_release -sr` == "18.04" ]; then true; else false; fi
$ echo $?
0
```

chapulina: Maybe what we need is to swap the true and false?

j-rivero: > Oh but I tried this locally on a Bionic machine and got this:

Was it under the usual /bin/bash or under /bin/sh? You can invoke the sh shell running `/bin/sh` and try running the command under that shell. By default Ubuntu uses /bin/bash where the `==` is valid.

chapulina: Oh I'm using `bash`, thanks for pointing it out.

j-rivero:
waiting for CI to start https://build.osrfoundation.org/job/ignition_msgs-ci-pr_any-ubuntu_auto-amd64/375

j-rivero:
Good to go: https://build.osrfoundation.org/job/ignition_msgs-ci-pr_any-ubuntu_auto-amd64/375

PR #177, opened 2020-04-23
ID closed branches to delete with 'hg heads -c'

scpeters:
Using 'hg log -r 'closed()' was inaccurate, as it includes
branches that were closed and then re-opened, causing the
git_delete_closed_hg_branches.bash script to delete open branches.
This improves the logic using 'hg heads -c'.

For gazebo:

~~~
# includes branches thet were closed and re-opened
$ hg log -r 'closed()' --template "{branch}\n" | sort | uniq | wc -l
    3144
~~~

~~~
# includes branches with a closed head and an open head
$ hg log -r 'closed() and head()' --template "{branch}\n" | sort | uniq | wc -l
    3124
~~~

~~~
# includes non-open branches only
$ hg branches -c | grep '(closed)$' | sed -e 's@ *[0-9]*[0-9]:.*@@' | sort | uniq | wc -l
    3121
~~~

the last one is the most accurate for this purpose

### DIFF
--- a/jenkins-scripts/tools/git_delete_closed_hg_branches.bash
+++ b/jenkins-scripts/tools/git_delete_closed_hg_branches.bash
@@ -9,7 +9,7 @@ echo =========== Done Cloning HG Repository
 echo
 echo =========== Identify Closed HG Branches
 CLOSED_BRANCHES=$(\
-  hg branches -c | grep '(closed)$' | sed -e 's@ *[0-9]*[0-9]:.*@@')
+  hg branches -c | grep '(closed)$' | awk '{ print $1 }')
 # the following doesn't work because 'hg log -r "closed()"' lists branches
 # that were closed and then re-opened:
 #  hg -R $HG_TMP log -r "closed()" --template '{branch}\n' | sort)
@@ -23,5 +23,4 @@ do
   echo remote git branch found: $b
   git push --delete origin $b
 done
-
 
--- a/jenkins-scripts/tools/git_delete_closed_hg_branches.bash
+++ b/jenkins-scripts/tools/git_delete_closed_hg_branches.bash
@@ -9,7 +9,13 @@ echo =========== Done Cloning HG Repository
 echo
 echo =========== Identify Closed HG Branches
 CLOSED_BRANCHES=$(\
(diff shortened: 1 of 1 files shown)


jenkins-scripts/tools/git_delete_closed_hg_branches.bash
```diff
@@ -9,7 +9,13 @@ echo =========== Done Cloning HG Repository
 echo
 echo =========== Identify Closed HG Branches
 CLOSED_BRANCHES=$(\
-  hg -R $HG_TMP log -r "closed()" --template '{branch}\n' | sort)
+  hg branches -c | grep '(closed)$' | sed -e 's@ *[0-9]*[0-9]:.*@@')
```
j-rivero, line 12: if the intention of that sed is to remove every but the first *column*, we could use `awk '{ print $1 }' which seems safer. No big deal anyway.
```suggestion
  hg branches -c | grep '(closed)$' | awk '{ print $1 }')
```
`

scpeters: yeah, that `awk` script is cleaner; I think I'm just haunted by the fact that [mercurial branches can have whitespace characters](http://mercurial.808500.n3.nabble.com/Escaping-characters-for-hg-update-td3994992.html) so I made a partial attempt to supporting that.

now that I think about it, the proper way would be to extract the commit hash from the `hg branches` call and use `hg log -r ${changeset} --template '"{branch}"'` and do escaping properly. I'm not sure we have any branch names with spaces though, what an evil thing that would be

jenkins-scripts/tools/git_delete_closed_hg_branches.bash
```diff
@@ -9,7 +9,13 @@ echo =========== Done Cloning HG Repository
 echo
 echo =========== Identify Closed HG Branches
 CLOSED_BRANCHES=$(\
-  hg -R $HG_TMP log -r "closed()" --template '{branch}\n' | sort)
+  hg branches -c | grep '(closed)$' | awk '{ print $1 }')
```
chapulina, line 12: Argh I merged too fast, sorry. It's missing the `-R $HG_TMP` here. See #178

PR #187, opened 2020-04-30
Move Github integration DSL code to configure block

j-rivero:
One of the problems to develop DSL code that we have now is that [the trick to generate jobs](https://github.com/jenkinsci/job-dsl-plugin/wiki/User-Power-Moves) (check syntax and results) does not work with the github PR plugin DSL code.

This PR changes the DSL to use configure block. Although we are losing a layer of indirection and almost inject code directly in the XML I think it worth to back to be able to generate code locally to speed up development.

I still need to configure test.dsl and a test repo to do the checks on Jenkins itself.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy
+++ b/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy
@@ -64,7 +64,7 @@ class GenericAnyJobGitHub
             cron()
             whiteListTargetBranches {
               'org.jenkinsci.plugins.ghprb.GhprbBranch' {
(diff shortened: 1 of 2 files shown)


j-rivero:
Using this branch to generate the test.dsl job and integrate it here: https://github.com/ignitionrobotics/testing/pull/1#issuecomment-622115859 (CI is failing because it tries to find tests, it should not do it after latest commit 8ed755c)

j-rivero:
New testing [Build Status](https://build.osrfoundation.org/job/_dsl_test/18/)

j-rivero:
I've detected a problem with a previous variable name, should be fixed in e7cf951

PR #188, opened 2020-04-30
Remove gazebo_ros_pkgs branch logic out of generic DSL classes

j-rivero:
I've removed all gazebo_ros_pkgs branches logic out of generic classes (it was duplicated inside `gazebo_ros_pkgs.dsl`).


chapulina:
Can we do a test run?

j-rivero:
Fixed some problems, success run https://build.osrfoundation.org/view/All/job/_dsl_gazebo_ros_pkgs/622/ . Don't want to improve it more since it's going to be replaced by pr #187 .

PR #194, opened 2020-05-07
Ubuntu CI: install both sdf8/9 for physics

scpeters:
@azeey noticed that we weren't installing sdformat9 in our Ubuntu CI:

* https://github.com/ignitionrobotics/ign-physics/pull/54#issuecomment-625319079

### DIFF
--- a/jenkins-scripts/lib/dependencies_archive.sh
+++ b/jenkins-scripts/lib/dependencies_archive.sh
@@ -659,6 +659,7 @@ IGN_PHYSICS_DEPENDENCIES="libbenchmark-dev \\
                           libignition-math6-dev \\
                           libignition-math6-eigen3-dev \\
                           libignition-plugin-dev \\
+                          libsdformat9-dev \\
                           libsdformat8-dev"
 IGN_PHYSICS_DART_FROM_PKGS="true"
 


jenkins-scripts/lib/dependencies_archive.sh
```diff
@@ -659,6 +659,7 @@ IGN_PHYSICS_DEPENDENCIES="libbenchmark-dev \\
                           libignition-math6-dev \\
                           libignition-math6-eigen3-dev \\
                           libignition-plugin-dev \\
+                          libsdformat9-dev \\
```
j-rivero, line 662: looks weird to me to have two versions of sdformat as dependencies listed together. I assume we are abusing the fact that we can co-install them to solve problems with different versions.

chapulina: Yeah I'd lean towards picking the SDF version based on `IGN_PHYSICS_MAJOR_VERSION`, but I approved because I thought there was an urgency.

scpeters:
> I'm assuming we never run ign-physics CI on Ubuntu versions below Bionic.

this is correct

jenkins-scripts/lib/dependencies_archive.sh
scpeters: > we are abusing the fact that we can co-install them to solve problems with different versions

this is correct; I just did it this way because it was more concise

if you like @j-rivero we can fix it in this PR or in a subsequent one

j-rivero: no big deal anyway.

scpeters: I've made the proper adjustment in #196

PR #197, opened 2020-05-08
Improve creation of new unofficial Gazebo ROS wrappers

j-rivero:
This is the first part of the improvements I made in the way of fully automatize the process of creating unofficial ROS wrappers. Here I include:

 * New script to get all debian/bloom metadata into the new repository created 
 * Improvements in the patching of names and change in the source package name
 * Safety checks
 * Update README


bloom/README
```diff
@@ -1,17 +1,24 @@
...
---------
+## Scripts:
+
+### Create a new gazeboX_rosY_pkgs repository
+
+For a new official wrappers the notation used below correspond to:
+`gazeboX_rosY_pkgs` (`X` is major version in Gazebo, `Y` is 1 or 2 for ROS)
+
+ 1. Create new repo under github.com/osrf/ organization named:
+    - https://github.com/osrf/gazeboX_rosY_pkgs-release
```
chapulina, line 9: I don't see any `ros1` repos. Should it say that `Y` is empty for ROS 1?

j-rivero:
> Does this doc need to be updated?

In fact we would need to have a single point of documentation. I'm fine with moving the instructions to there and remove the README from here and the internal wiki

chapulina:
> we would need to have a single point of documentation

+1 to that. Ideally we need it in a public place, I don't have a preference about where. If we remove it from `gazebo_ros_pkgs`, we should at least leave a link to the new location.

bloom/README
j-rivero: 098fc2c

j-rivero:
> > 
> 
> +1 to that. Ideally we need it in a public place, I don't have a preference about where. If we remove it from `gazebo_ros_pkgs`, we should at least leave a link to the new location.

I don't see the RELEASE file in `noetic-devel` nor in `ros2` branch. The point of having it here is that it is close to the scripts needed.

j-rivero:
I've imported the release info from `gazebo_ros_pkgs`, I'm going to modify the `melodic-devel` and `noetic-devel` branches to point to this doc here. Merging, don't want to lose the PR in the limbo again :)

PR #208, opened 2020-05-31
brew: decode bottle filename as URI.

scpeters:
I noticed that the dartsim@6.10.0 bottle filename had changed the `@` character to `%40` before upload to s3, which caused it to fail to download during `brew install dartsim@6.10.0` (see the build artifacts in https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/830/label=osx_highsierra/ ).

This is caused by https://github.com/Homebrew/brew/pull/7532 which encodes the filename in json. This PR decodes it, which should fix the problem.


scpeters:
I noticed this problem in the following build:

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/label=osx_mojave/832/) https://build.osrfoundation.org/job/generic-release-homebrew_bottle_builder/832/label=osx_mojave/

I manually copied the dartsim bottles on s3 to versions with the correct name, and then that build was successful. This commit will allow the `dartsim@6.10.0` bottle to be correctly named the next time it is built.

PR #220, opened 2020-06-11
Use tinyxml2 for sdformat10

mjcarroll:
Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/jenkins-scripts/lib/dependencies_archive.sh
+++ b/jenkins-scripts/lib/dependencies_archive.sh
@@ -55,15 +55,13 @@ BREW_BASE_DEPENDCIES="mercurial git cmake"
 # 1. SDFORMAT
 # ruby for xml_schemas generation and libxml2-utils for xmllint used in tests
 SDFORMAT_NO_IGN_DEPENDENCIES="${pythonv}     \\
-                              libxml2-utils"
+                              libxml2-utils  \\
+                              libtinyxml-dev"
 
 # SDFORMAT 10 and above use tinyxml2
 if [[ ${SDFORMAT_MAJOR_VERSION} -ge 10]]; then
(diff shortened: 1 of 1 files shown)


mjcarroll:
Used for https://github.com/osrf/sdformat/pull/264

j-rivero:
We also need:
 * Brew: https://github.com/osrf/homebrew-simulation/pull/1054
 * Windows: the changes in sdformat PR remove the tinyxml embedded code. That needs to be fixed before merging.

jenkins-scripts/lib/dependencies_archive.sh
```diff
@@ -55,8 +55,16 @@ BREW_BASE_DEPENDCIES="mercurial git cmake"
...
 # ruby for xml_schemas generation and libxml2-utils for xmllint used in tests
 SDFORMAT_NO_IGN_DEPENDENCIES="${pythonv}     \\
-                              libxml2-utils  \\
-                              libtinyxml-dev"
+                              libxml2-utils"
+
+# SDFORMAT 10 and above use tinyxml2
+if [[ ${SDFORMAT_MAJOR_VERSION} -ge 10]]; then
+  SDFORMAT_NO_IGN_DEPENDENCIES="${SDFORMAT_NO_IGN_DEPENDENCIES} \\
+                                libtinyxml2-dev"
```
scpeters, line 63: this is what it should look like eventually, but until everything is merged, I think it will keep other branches targeting `sdf10` working if we install both `libtinyxml-dev` and `libtinyxml2-dev` for 10+. Once https://github.com/osrf/sdformat/pull/264 is merged, the `libtinyxml-dev` line can be removed.

how does that sound?

mjcarroll: Completely reasonable, updated.

PR #226, opened 2020-06-17
Clean up deprecated chroot scripts + bitbucket integration

j-rivero:
The PR removes old linux related script that were hosted at `release-tools/jenkins-scripts/{lib}` . I've included some old dsl files. As far as I can say none of them is in use right now.


scpeters:
I guess we'll find out!

PR #228, opened 2020-06-25
Migrate VRX, SubT and gzdev to Github

j-rivero:
//cc @caguero , @nkoenig 

Migrate the VRX, Subt and gzdev CI code to use github. These should be the last projects we change to the new CI integration.

### DIFF
--- a/jenkins-scripts/dsl/subt.dsl
+++ b/jenkins-scripts/dsl/subt.dsl
@@ -169,7 +169,7 @@ all_supported_distros.each { distro ->
         git {
           remote {
             github("osrf/subt", 'https')
-            branch('${BRANCH}')
+            branch('master')
           }
 
           extensions {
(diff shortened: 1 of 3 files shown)


jenkins-scripts/dsl/subt.dsl
```diff
@@ -166,9 +166,17 @@ all_supported_distros.each { distro ->
     long_run_job.with
     {
       scm {
-        hg('https://bitbucket.org/osrf/subt') {
-          branch('default')
-          subdirectory('subt')
+        git {
+          remote {
+            github("osrf/subt", 'https')
+            branch('${BRANCH}')
```
scpeters, line 172: the old one is always running on `default`; should this be `master` instead of `${BRANCH}`?

j-rivero: good catch! fixed in ce991a6

PR #229, opened 2020-06-25
Remove bitbucket dsl

### LINKED ISSUE #216: vrx, subt, and servicesim CI scripts still using bitbucket
* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/vrx.dsl#L50-L77

* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/subt.dsl#L92
* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/subt.dsl#L127
* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/subt.dsl#L169
* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/subt.dsl#L203-L208

* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/servicesim.dsl#L60-L77
* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/servicesim.dsl#L114

j-rivero:
Requires merging https://github.com/ignition-tooling/release-tools/pull/228 first. Fixes #216 

Remove old classes in DSL that were duplicated for Github and the `GITHUB` parameter no longer needed.


chapulina:
Just a drive-by comment that ARIAC is still on BitBucket (it uses git).

:car: 💨 Bye!

j-rivero:
> Just a drive-by comment that ARIAC is still on BitBucket (it uses git).

uh, do we need to maintain ariac active in the scripts?

sloretz:
> uh, do we need to maintain ariac active in the scripts?

I think it would be fine to remove any ARIAC jobs. I don't think we'll need to make another release.

jenkins-scripts/docker/lib/generic-abi-base.bash
```diff
@@ -68,13 +68,7 @@ echo '# BEGIN SECTION: compile and install branch: ${DEST_BRANCH}'
...
 chown -R root:root /tmp/${ABI_JOB_SOFTWARE_NAME}
 cd /tmp/${ABI_JOB_SOFTWARE_NAME}
-if ${GITHUB}; then
-  git fetch origin
-  git checkout origin/${DEST_BRANCH}
-else
-  hg pull
-  hg up ${DEST_BRANCH}
-fi
+git checkout origin/${DEST_BRANCH}
```
scpeters, line 71: the `git fetch origin` line is removed; is that intentional?

scpeters (commented):
the changes look ok, though a lot of files are deleted. I'd like to see this branch tested with https://build.osrfoundation.org/job/_dsl_generate_all_dsls/ before merging

jenkins-scripts/docker/lib/generic-abi-base.bash
j-rivero: it was not!  47a364 . Good catch

scpeters:
> the changes look ok, though a lot of files are deleted. I'd like to see this branch tested with https://build.osrfoundation.org/job/_dsl_generate_all_dsls/ before merging

@j-rivero what do you think of testing this branch with all the dsl jobs?

j-rivero:
> @j-rivero what do you think of testing this branch with all the dsl jobs?

Probably the only danger is the GITHUB variable, defined in [current code in DSL](https://github.com/ignition-tooling/release-tools/pull/229/files#diff-6d23621bef390895cdcfe6d0156ffd97L31 ) being used in [some checkouts](https://github.com/ignition-tooling/release-tools/pull/229/files#diff-aecce6fa8f122c829c7fade8888f8cadL258). The DSL code will run with this branch but on production jobs will run without this branch. If I'm not wrong, the default value in current code to True should make things to work. We can give it a try.

scpeters:
oh, I guess we can't run the following job with a branch parameter:

* https://build.osrfoundation.org/job/_dsl_generate_all_dsls/637/

I guess we can just try merging it and see what happens

scpeters:
> oh, I guess we can't run the following job with a branch parameter:
> 
> * https://build.osrfoundation.org/job/_dsl_generate_all_dsls/637/
> 
> I guess we can just try merging it and see what happens

never mind, I see that this will break the subt dsl:

* https://github.com/ignition-tooling/release-tools/blob/remove_bitbucket_dsl/jenkins-scripts/dsl/subt.dsl#L246

we need to take a closer look at all of these changes or get a way to test this branch

j-rivero:
> > 
> 
> never mind, I see that this will break the subt dsl:

wow, good catch!

> we need to take a closer look at all of these changes or get a way to test this branch

I've [generated](https://github.com/jenkinsci/job-dsl-plugin/wiki/User-Power-Moves#generate-a-job-configxml-without-having-to-fire-up-jenkins) all our .dsl jobs in my local system using `for f in $(ls *.dsl); do jentest $f || break ; done` (being jentest an alias to the local generation) on the dsl directory, nothing seems to be still broken in the package generation.

scpeters:
> I've [generated](https://github.com/jenkinsci/job-dsl-plugin/wiki/User-Power-Moves#generate-a-job-configxml-without-having-to-fire-up-jenkins) all our .dsl jobs in my local system using `for f in $(ls *.dsl); do jentest $f || break ; done` (being jentest an alias to the local generation) on the dsl directory, nothing seems to be still broken in the package generation.

ok, I'll take one more look then

scpeters (approved):
looks great! I just updated some comments in eec66cef09b86af9f8620dad276449d978c6b751 and it looks ready to me

j-rivero:
> looks great! I just updated some comments in [eec66ce](https://github.com/ignition-tooling/release-tools/commit/eec66cef09b86af9f8620dad276449d978c6b751) and it looks ready to me

Thanks Steve, let's merge today and see during the weekend if something is broken.

PR #230, opened 2020-06-26
Integrate ABI jobs in Github

### LINKED ISSUE #191: ABI jobs into Github
Integrate the ABI jobs into Github in the same way as the other 3 platforms jobs. Currently they are being run from `-ci-pr_any-` from linux but the status is not being published to the PRs.

j-rivero:
The pull request change how ABI jobs are triggered:
  * Before, it was a conditional step in compilation jobs if target branch was different than master
  * After: the abi job integrates itself in github calls only if the target branch is a supported branch (major version branches)

I'm leaving the same parameters as before to be able to run builds manually in the same way.

**Testing:**

Be careful to run dsl changes on Jenkins since they will affect production code. I've used my fork to check:
 * Test PR with ABI breaking against sdf9 https://github.com/j-rivero/sdformat/pull/1
 * Test PR with no ABI breaking changes in sdf8 https://github.com/j-rivero/sdformat/pull/2

All DSL runs fine on my local system and the produced xml for branches seems fine to me.

Closes #191

### DIFF
--- a/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy
+++ b/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy
@@ -26,12 +26,6 @@ class GenericAnyJobGitHub
     // Get repo name for relativeTargetDirectory
     String github_repo_name = github_repo.substring(github_repo.lastIndexOf("/") + 1)
 
-    // Transform GStringImp into real string
-    ArrayList supported_branches_str = []
-    supported_branches.each { branch ->
-      supported_branches_str.add(branch.toString())
-    }
-
     job.with
     {
       parameters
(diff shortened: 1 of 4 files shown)


scpeters:
I just tested the with https://github.com/osrf/sdformat/pull/301, and it's not working:

[Build Status](https://build.osrfoundation.org/job/sdformat-abichecker-any_to_any-ubuntu_auto-amd64/834/) https://build.osrfoundation.org/job/sdformat-abichecker-any_to_any-ubuntu_auto-amd64/834/

scpeters:
it looks like that job doesn't have the `GITHUB` variable set, but it also isn't setting the `SRC_BRANCH` or `DEST_BRANCH` variables

j-rivero:
> it looks like that job doesn't have the `GITHUB` variable set, 

Ouch, my bad sorry. I've merge #223 to help fixing it.

> but it also isn't setting the `SRC_BRANCH` or `DEST_BRANCH` variables

I think it is a cosmetic thing for the build descriptions and should not stop of the build to be complete, although I need to fix it.

Seems to work now: [Build Status](https://build.osrfoundation.org/job/sdformat-abichecker-any_to_any-ubuntu_auto-amd64/835/)

scpeters:
thanks for the quick fix!

scpeters:
I can't figure out why the following comment didn't start an ign-physics ABI checker job:

* https://github.com/ignitionrobotics/ign-physics/pull/56#issuecomment-652089154

PR #231, opened 2020-06-29
Remove old scripts: catkin-debs, ariac, eigen, and manual-jenkins-scripts

j-rivero:
Found more not in use code in the repo:
 * Old scripts for ariac or testing eigen
 * catlin-debs tools imported from ROS buildfarm for chroots
 * Remove chroots from manual jenkins scripts (designed to local test changes)


scpeters:
if we need to revive ariac, we can bring these scripts back; it seems fine to delete them

PR #235, opened 2020-07-06
Fix ABI jobs integration on ignition CI

j-rivero:
There was a mistake in a parameter when calling `OSRFLinuxABIGitHub` using checkout subir instead of repository qualified name. This PR should fix the problem and also add correct XML tags for permissions.

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -222,9 +222,9 @@ ignition_software.each { ign_sw ->
       abi_job_names[ign_sw] = "ignition_${ign_sw}-abichecker-any_to_any-ubuntu_auto-${arch}"
       def abi_job = job(abi_job_names[ign_sw])
       checkout_subdir = "ign-${ign_sw}"
-      OSRFLinuxABIGitHub.create(abi_job, checkout_subdir)
-      GenericAnyJobGitHub.create(abi_job,
-                        "ignitionrobotics/ign-${ign_sw}",
+      repo = "ignitionrobotics/ign-${ign_sw}"
+      OSRFLinuxABIGitHub.create(abi_job, repo)
+      GenericAnyJobGitHub.create(abi_job, repo,
                         all_branches(ign_sw) - [ 'master'])
       abi_job.with
       {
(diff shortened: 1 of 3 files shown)


jenkins-scripts/dsl/ignition.dsl
```diff
@@ -222,9 +222,9 @@ ignition_software.each { ign_sw ->
       abi_job_names[ign_sw] = "ignition_${ign_sw}-abichecker-any_to_any-ubuntu_auto-${arch}"
       def abi_job = job(abi_job_names[ign_sw])
       checkout_subdir = "ign-${ign_sw}"
-      OSRFLinuxABIGitHub.create(abi_job, checkout_subdir)
-      GenericAnyJobGitHub.create(abi_job,
-                        "ignitionrobotics/ign-${ign_sw}",
+      repo = "ignitionrobotics/ign-${ign_sw}"
+      OSRFLinuxABIGitHub.create(abi_job, repo)
```
scpeters, line 226: I'm not sure why the second argument is provided to `OSRFLinuxABIGitHub.create` here, since only `abi_job` is passed in [gazebo.dsl](https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/gazebo.dsl#L92) and [sdformat.dsl](https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/sdformat.dsl#L40)

scpeters:
testing:

* [Build Status](https://build.osrfoundation.org/job/_dsl_ignition/997/) https://build.osrfoundation.org/job/_dsl_ignition/997/

scpeters:
testing with https://github.com/ignitionrobotics/ign-physics/pull/56: [Build Status](https://build.osrfoundation.org/job/ignition_physics-abichecker-any_to_any-ubuntu_auto-amd64/429/) https://build.osrfoundation.org/job/ignition_physics-abichecker-any_to_any-ubuntu_auto-amd64/429/

scpeters:
> testing with [ignitionrobotics/ign-physics#56](https://github.com/ignitionrobotics/ign-physics/pull/56): [Build Status](https://build.osrfoundation.org/job/ignition_physics-abichecker-any_to_any-ubuntu_auto-amd64/429/) https://build.osrfoundation.org/job/ignition_physics-abichecker-any_to_any-ubuntu_auto-amd64/429/

build failed due to a compilation error, but it was triggered correctly, so that's what matters

jenkins-scripts/dsl/ignition.dsl
j-rivero: The method has an overloaded `create` that took two parameters and setup github checkout before calling the `create` used in gazebo and sdformat. I've removed that logic in 575590f to simplify the use and don't fall in these kind of errors again. The class should be renamed after #229 since it has nothing to do with GitHub.

PR #238, opened 2020-07-09
Run vcs pull after vcs import to get latest changes

j-rivero:
While testing some changes in ign-cmake I realize that the CI job preserve the workspace and `vcs import` does not update the checkout directories if they exists. Adding a `vcs pull` updates all repositories to latest commits.

Testing it here: [Build Status](https://build.osrfoundation.org/view/All/job/ign_gui-pr-win/328/)


chapulina:
> the CI job preserve the workspace 

Another problem that could come from this is mixing distributions. Maybe it's best to completely clear the workspace? Otherwise  `colcon build` may build several distributions at the same time.

An alternative solution would be to build `--packages-up-to` the correct package, not sure if this is already part of `%COLCON_EXTRA_ARGS%`, maybe?

chapulina:
Merging so we can fix some CI. My other comment can be addressed later if needed.

PR #243, opened 2020-07-18
Fix sdformat10 debbuilder job

scpeters:
It needs to be defined in sdformat.dsl, not ignition.dsl because otherwise it makes the job like this:

* https://build.osrfoundation.org/job/ign-sdformat10-debbuilder/

Replace sdformat7-debbuilder with sdformat10-debbuilder, since we aren't planning to make sdformat7 releases.

Don't add sdformat10 to supported branches yet because we don't need ABI checking.

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -58,7 +58,6 @@ ignition_debbuild = ignition_software + [ 'fuel-tools5',
                                           'physics3',
                                           'rendering4',
                                           'sensors4',
-                                          'sdformat10',
                                           'transport9']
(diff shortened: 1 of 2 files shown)


scpeters:
> we would need to get rid of the gz11 denomination at some point :)

I think once `libsdformat6` is no longer supported, the logic will not be necessary. We could rename the variables to refer to `c++17` instead of `gz11` if we wanted to clarify the purpose of the logic

j-rivero:
> I think once `libsdformat6` is no longer supported, the logic will not be necessary. We could rename the variables to refer to `c++17` instead of `gz11` if we wanted to clarify the purpose of the logic

that was exactly my though when I was reviewing the DSL one week ago.

PR #245, opened 2020-07-21
Avoid lintian in debbuilds on ARM (hangs because syscall not implemented in qemu)

j-rivero:
Lintian (QA checker for deb packages) on ARM on Focal hangs qemu (missing syscall). This PR adds a couple of workarounds:

 * debbuild supports `--no-lintian` flag
 * Did not find the same for gbp so ... kill the whole lintian app by using /bin/true.


jenkins-scripts/docker/lib/debbuild-base.bash
```diff
@@ -183,7 +183,14 @@ echo '# END SECTION'
 fi
 
 echo '# BEGIN SECTION: create source package' \${OSRF_VERSION}
-debuild --no-tgz-check -uc -us -S --source-option=--include-binaries
+
+# lintian triggers a problem on arm in Focal when using qemu, avoid it
+no_lintian_param=""
+if [[ ${DISTRO} == 'focal' && ${ARCH} == 'arm64' ]]; then
```
scpeters, line 189: an armhf build is hanging too:

* https://build.osrfoundation.org/job/gazebo9-debbuilder/587/console

PR #249, opened 2020-07-24
Support to use ignition_version to release ign_ros_bridge

j-rivero:
Add a new parameter called `IGNITION_COLLECTION` to support builds of `ros_ign` with custom ignition releases.

DSL is tested here: [Build Status](https://build.osrfoundation.org/job/_dsl_ros1_ign_bridge/29/)

Output of release script displays correctly the use of the parameter and the inclusion in the Jenkins call:
```
jrivero@nium bloom $ (ign_bridge) ./ros_ign_bridge-release.py.bash 1.0.0 foo-release melodic pass --ignition-collection blueprint -r 1 --dry-run
Accessing: http://build.osrfoundation.org/job/ros-ign-image-bloom-debbuilder/buildWithParameters?PACKAGE=ros-ign-image&UPLOAD_TO_REPO=stable&RELEASE_VERSION=1&UPSTREAM_RELEASE_REPO=foo-release&token=pass&VERSION=1.0.0&IGNITION_COLLECTION=blueprint&ARCH=amd64&DISTRO=bionic&ROS_DISTRO=melodic
```

### DIFF
--- a/bloom/release-bloom.py
+++ b/bloom/release-bloom.py
@@ -6,7 +6,7 @@ import sys
 import urllib
 import argparse
 
-USAGE = 'release-bloom.py <package> <version> <upstream_release_repo> <ros_distro> [--ignition-collection collection_name] <jenkinstoken>'
+USAGE = 'release-bloom.py <package> <version> <upstream_release_repo> <ros_distro> [--ignition-version version_name] <jenkinstoken>'
 JENKINS_URL = 'http://build.osrfoundation.org'
 JOB_NAME_PATTERN = '%s-bloom-debbuilder'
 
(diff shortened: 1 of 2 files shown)


bloom/release-bloom.py
```diff
@@ -77,6 +79,9 @@ def call_jenkins_jobs(argv):
         params['ROS2'] = True
         ubuntu_per_ros_distro = UBUNTU_DISTROS_IN_ROS2
 
+    if args.ignition_collection:
+        params['IGNITION_COLLECTION'] = args.ignition_collection
```
chapulina, line 83: Maybe we should use `IGNITION_VERSION` everywhere to avoid confusion?

j-rivero: changed

jenkins-scripts/dsl/ros1_ign_bridge.dsl
```diff
@@ -33,7 +33,8 @@ bridge_packages.each { pkg ->
     }
 
     parameters {
-      stringParam("PACKAGE","$pkg_dashed","Package name to be built")
+        stringParam("PACKAGE","$pkg_dashed","Package name to be built")
+        stringParam("IGNITION_VERSION", '', 'Ignition release supported in the binaries')
```
chapulina, line 37: Would it be a good idea to have a default value here?

j-rivero: I don't think that we have a good candidate given that it will vary with time and we probably want it to fail if unset instead of running a wrong distro.

PR #255, opened 2020-08-03
More QML dependencies for ign-gui

chapulina:
We're missing some QML dependencies, which is causing test failures on https://github.com/ignitionrobotics/ign-gui/pull/69/. 

Test build using this branch queued: [Build Status](https://build.osrfoundation.org/job/ignition_gui-ci-pr_any-ubuntu_auto-amd64/567/)


chapulina:
It fixed the tests, merging

PR #262, opened 2020-08-08
Workaround for manlib denied problem

j-rivero:


j-rivero:
Code is failing in jobs using `DOCKER_IN_DOCKER` with the an apparmor error that I can reproduce locally:

```
jrivero@nium ~ $ docker run --privileged --security-opt apparmor=docker-unconfined -e DISPLAY=unix:0 -v /sys:/sys:ro -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/.X11-unix:/tmp/.X11-unix:rw -v /run/log:/run/log:ro -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v /proc/  --device /dev/snd --tty --rm ubuntu:xenial /bin/bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"apply apparmor profile: apparmor failed to apply profile: open /proc/self/attr/exec: no such file or directory\"": unknown.
```

My impression is that most of errors in gazebo9-debbuilder come from the problem with `/dev/snd` pull #263 . I'm building the failing packages again to confirm this.

j-rivero:
problems with infrastructure are solved by #264

PR #265, opened 2020-08-11
Do not fail uploades of source packages

j-rivero:
Some nightly packages have problems uploading the source package since they conflict between distributions. While is not a critical problem itself (few use cases for source packages) it makes the uploads to fail so no binaries go to the repo which in fact it is a critical problem. I'm making the script not to fail but report unstable until I code the right fix for the whole thing.

Tested:

 * Failing [Build Status](https://build.osrfoundation.org/view/All/job/repository_uploader_ng/14910/)
 * Good (although the parser was not properly setup) [Build Status](https://build.osrfoundation.org/view/All/job/repository_uploader_ng/14913/)


j-rivero:
> Can we ticket an issue for this?

sure, ready #267

PR #272, opened 2020-08-22
👩‍🌾 Foxy and Rolling supported, Galactic staged

chapulina:
* Make Foxy and Rolling supported
* Set Galactic as the next one coming up

### DIFF
--- a/jenkins-scripts/dsl/_configs_/Globals.groovy
+++ b/jenkins-scripts/dsl/_configs_/Globals.groovy
@@ -18,7 +18,9 @@ class Globals
                      'noetic'   : ['focal'] ,
                      'dashing'  : ['bionic'] ,
                      'eloquent' : ['bionic'] ,
-                     'foxy'     : ['focal']]
+                     'foxy'     : ['focal'] ,
+                     'galactic' : ['focal'] ,
+                     'rolling'  : ['focal']]
 
    // This should be in sync with archive_library
    static gz_version_by_rosdistro = [ 'kinetic'  : ['7'] ,
@@ -26,7 +28,9 @@ class Globals
                                       'noetic'   : ['11'] ,
                                       'dashing'  : ['9'] ,
                                       'eloquent' : ['9'] ,
-                                      'foxy'     : ['11']]
+                                      'foxy'     : ['11'] ,
+                                      'galactic' : ['11'] ,
+                                      'rolling'  : ['11']]
 
    static ArrayList get_ros_distros_by_ubuntu_distro(String ubuntu_distro)
    {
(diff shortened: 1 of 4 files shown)


jenkins-scripts/dsl/_configs_/Globals.groovy
```diff
@@ -102,11 +106,11 @@ class Globals
 
    static ArrayList get_ros2_suported_distros()
    {
-     return [ 'dashing', 'eloquent' ]
+     return [ 'dashing', 'eloquent', 'foxy', 'rolling' ]
    }
 
    static String get_ros2_development_distro() {
-     return 'foxy'
+     return 'galactic'
```
mjcarroll, line 113: Should this be 'rolling'?

scpeters:
I'll defer to @mjcarroll's review of this

jenkins-scripts/dsl/_configs_/Globals.groovy
chapulina: Yeah maybe this entire logic could be a bit updated to account for the new rolling paradigm. The reason I didn't put `rolling` here is that we probably want rolling install jobs. See


https://github.com/ignition-tooling/release-tools/blob/c5045684ef7915bc29c6573ad4a624de5b402084/jenkins-scripts/dsl/gazebo_ros_pkgs.dsl#L165-L168

mjcarroll: Yes, it probably makes sense to have install jobs.  I think now generally that the ordering may be a bit inverted, in that we will have packages for rolling, that will eventually become the packages for galactic and further.

chapulina: Cool, so I think `galactic` here is ok for now?

j-rivero: agree with Louise. The logic probably needs to be updated to match what currently is the "developement_distro".

j-rivero (approved):

j-rivero:
Doing a couple of tests before merging:
 *  [Build Status](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-default_foxy-focal-amd64/21/)
 * [Build Status](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-default_rolling-focal-amd64/2/)

jenkins-scripts/dsl/_configs_/Globals.groovy
mjcarroll: :+1:

mjcarroll (approved):

j-rivero:
> Doing a couple of tests before merging:

I think we need to take care of defining the right branches for foxy and rolling somewhere. This change would create `gazebo_ros_pkgs` jobs that will fail.

chapulina:
> we need to take care of defining the right branches for foxy and rolling somewhere.

They should both be the `ros2` branch at the moment. I guess it is looking for `foxy` and `rolling`

j-rivero:
After a5dd2ca 

 * rolling [Build Status](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-default_rolling-focal-amd64/3/)
 * foxy [Build Status](https://build.osrfoundation.org/job/ros2_gazebo_pkgs-ci-default_foxy-focal-amd64/22/)

Seems like there are problems with the tests finding plugins but the compilation mostly work and I think the changes in this PR are good to go.

PR #277, opened 2020-09-15
Replace old packages uploader

j-rivero:
This change replace the use of the old `repository_uploader_ng` that uploaded packages to build.osrfoundation.org to `repository_uploader_packages` that uploads packages directly to packages.osrfoundation.org. Currently the first one was forwarded calls to the second one which is the most important.

This will kill the old.packages.osrfoundation.org, announced long ago [here](https://community.gazebosim.org/t/packages-osrfoundation-org-new-machine/426/3?u=jrivero)


j-rivero:
Tested:
 * DSL: https://build.osrfoundation.org/view/All/job/_dsl_ignition/1021/
 * Use repository_uploader_packages directly with fuel-tools: https://build.osrfoundation.org/job/repository_uploader_packages/6507/

scpeters (approved):

PR #279, opened 2020-09-16
Export Dockerfile/build.sh as artifacts from each build 

j-rivero:
This should help people to reproduce the Jenkins builds locally. The first commit is the only one interesting https://github.com/ignition-tooling/release-tools/commit/e2b17390fbf405b471daba6403929c815d8e5e48, second is just proper tabs.

Testing in DSL here: [Build Status](https://build.osrfoundation.org/view/All/job/_dsl_sdformat/912/)


scpeters (approved):
tested here: https://build.osrfoundation.org/view/All/job/sdformat-ci-sdformat9-bionic-amd64/39/

looks good

Issue #288, opened 2020-09-23
Enable GCC8 on ign-gazebo deb builds


chapulina:
`ign-gazebo4` Bionic deb builds are [failing](https://build.osrfoundation.org/job/ign-gazebo4-debbuilder/113/console) to complete because they can't find filesystem when compiling a test:

```
/var/lib/jenkins/workspace/ign-gazebo4-debbuilder/build/ignition-gazebo-4.0.0~pre1/test/integration/log_system.cc:24:10: fatal error: filesystem: No such file or directory
 #include <filesystem>
          ^~~~~~~~~~~~
```

I believe we could solve it on the source code by using `experimental::filesystem` if GCC < 8. But is there a reason not to enable GCC8 on the debbuilds?

I was able to [successfully build](https://build.osrfoundation.org/job/ign-gazebo4-debbuilder/119/) `4.0.0~pre1` for Bionic using this patch https://github.com/ignition-tooling/release-tools/commit/429d990ab48b8b04989bf27ab4dfafbddd4ba82f.

Can we just set `NEED_C17_COMPILER` to true for these builds?

j-rivero:
Using the non default compiler for our sources is a not-very-nice practice since it blocks people to compile our code using default options. It is not a blocker since the configuration can be done easily using official Ubuntu/Debian packages.

For the debbuilders: using a non-default compiler (like in this case) is a valid case (for me) as far as it does not require to change system configuration to the users of those binaries, which I think is this situation.

The right fix for the problem is to inject the `NEED_C17_COMPILER` in the DSL code only when generating the ign-gazebo4 job. @chapulina I can take care of it if you don't have time to deal with DSL code.

chapulina:
@j-rivero , maybe #289 does it? I didn't do that directly because I didn't want to re-run the DSL.

PR #289, opened 2020-09-23
Enable GCC8 for ign-gazebo deb builds

chapulina:
Possible solution to #288 

Copied the approach from SDF:

https://github.com/ignition-tooling/release-tools/blob/35e122906ccf7b6cff8623a82cd150bf5b364386/jenkins-scripts/dsl/sdformat.dsl#L275-L282


j-rivero:
These changes are totally right to fix the bug. Since I don't have a good reason to keep two variables for the same thing, I've created #290 as an alternative to solve the two problems at the same time.

PR #294, opened 2020-09-30
Clean up installspace + buildspace in ROS workspaces

j-rivero:
The PR should remove files in build and install spaces created by ROS. This should help to reduce the disk space used by the different builds.

### DIFF
--- a/jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
+++ b/jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
@@ -12,6 +12,7 @@ fi
 
 [[ -z ${USE_GZ_VERSION_ROSDEP} ]] && USE_GZ_VERSION_ROSDEP=false
 [[ -z ${USE_CATKIN_MAKE} ]] && USE_CATKIN_MAKE=false # use colcon by default
+[[ -z ${KEEP_WS_FILES} ]] && KEEP_WS_FILES=false # clean up install/ build/
 
 if ${USE_CATKIN_MAKE}; then
   export CMD_CATKIN_CONFIG=""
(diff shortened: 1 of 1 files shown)


jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
```diff
@@ -123,6 +124,12 @@ for d in \$DIRS; do
 done
 echo '# END SECTION'
 
+# clean up build and install directories to save disk space in Jenkins server
+if [[ ! $KEEP_WS_FILES ]]; then
+  rm -fr ${CATKIN_WS}/install/*
+  rm -fr ${CATKIN_WS}/build/*
```
chapulina, line 130: Do we need to remove `devel` when using catkin?

scpeters: `colcon` makes a `log` folder too

j-rivero: This  is subt-pr_any- job in drogon:
```
   1.4 GiB [##########] /install
  139.9 MiB [          ] /build
   88.2 MiB [          ] /devel
    4.0 KiB [          ] /src
    4.0 KiB [          ]  .catkin_workspace
```
and ros2-ci-pr_any-foxy
```                         /..                                                                                                                                                    
  516.0 MiB [##########] /build
    6.4 MiB [          ] /install
    5.1 MiB [          ] /log
    8.0 KiB [          ] /src
```
There is at least one order of magnitude of difference between install/build and the rest. No problem on including more things.

scpeters (approved):

jenkins-scripts/docker/lib/_ros_setup_buildsh.bash
scpeters: that's fair; I've approved as is

PR #304, opened 2020-10-02
Improve free of disk space: reduce limits and clean /build dirs

j-rivero:
There are problems related to nodes disk space since sometimes builds are so big (gazebo and/or subt can go over 10Gb) that the disk space available is not enough. Other times the Jenkins configuration disables the node under 10Gb of free space. For trying to avoid that situation the PR does two thigns:

 * Reduce the percents of the disk usage to trigger cleaning.
 * Add a new method to find build/ directories bigger than 5Gb (gazebo packages are under this) as a final resource. These directories are from time to time left when a build failed and in the case of Gazebo can be up to 14Gb.

### DIFF
--- a/jenkins-scripts/docker/lib/boilerplate_prepare.sh
+++ b/jenkins-scripts/docker/lib/boilerplate_prepare.sh
@@ -20,7 +20,7 @@ if [[ -z ${DO_NOT_CHECK_DOCKER_DISK_USAGE} ]]; then
     docker_device=$(df '/var/lib/docker' | awk '{ print $1 }' | tail -n 1)
     # in seconds: 5 days = 432000s
     PERCENT_DISK_USED=$(df -h ${docker_device} | grep ${docker_device} | sed 's:.* \([0-9]*\)%.*:\1:')
-    if [[ $PERCENT_DISK_USED -gt 90 ]]; then
+    if [[ $PERCENT_DISK_USED -gt 70 ]]; then
         echo "Space left is low: ${PERCENT_DISK_USED}% used"
         echo "Run docker cleaner !!"
         wget https://raw.githubusercontent.com/spotify/docker-gc/master/docker-gc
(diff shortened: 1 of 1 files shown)


jenkins-scripts/docker/lib/boilerplate_prepare.sh
```diff
@@ -50,6 +50,27 @@ if [[ -z ${DO_NOT_CHECK_DOCKER_DISK_USAGE} ]]; then
           [[ -n $(sudo docker images -a -q) ]] && sudo docker rmi $(sudo docker images -a -q) || true
         fi
     fi
+
+    # if not enough, try to clean up build/ directories
+    PERCENT_DISK_USED=$(df -h ${docker_device} | grep ${docker_device} | sed 's:.* \([0-9]*\)%.*:\1:')
+    if [[ $PERCENT_DISK_USED -gt 85 ]]; then
+        MAX_SIZE_FOR_BUILD_DIRS="5G"
+        CMD_FIND_BIG_DIRS=(find ${HOME}/workspace -name build -exec du -h -d 0 -t ${MAX_SIZE_FOR_BUILD_DIRS} {} \;)
```
scpeters, line 58: is there a missing `$` after `=`? i.e. `=(` vs `=$(`

(`$` looks like a funny mustache)

scpeters: oh, maybe not; I am still a bash padawan

nuclearsandwich: Yeah this is setting the variable to a Bash array which is being used later on https://github.com/ignition-tooling/release-tools/pull/304/files#diff-0846c9d3ff0a72e925d87240401660f255512aa7a9eb532b39657209792ba5aeR62

nuclearsandwich (approved):

jenkins-scripts/docker/lib/boilerplate_prepare.sh
nuclearsandwich: Quoting it makes it an array with a single element instead of each space-separated word its own.

scpeters (approved):

PR #315, opened 2020-10-08
Disable CI for ign-rndf

scpeters:
ign-rndf CI is failing as it doesn't have a main branch, but it has been inactive for a while; I propose to disable its CI jobs.

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -13,7 +13,6 @@ ignition_software = [ 'cmake',
                       'physics',
                       'plugin',
                       'rendering',
-                      'rndf',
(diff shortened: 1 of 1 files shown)


chapulina (approved):
@caguero , should we archive https://github.com/ignitionrobotics/ign-rndf/ ?

caguero:
I think it's OK to disable CI, but I'd leave the repo available just in case someone is using it out there.

scpeters:
> I think it's OK to disable CI, but I'd leave the repo available just in case someone is using it out there.

archived repositories can still be cloned and forked; I think the main thing missing is being able to open and comment on issues

scpeters:
I'll merge this now, but discussion can continue about the archiving question

PR #316, opened 2020-10-08
Revert "Dome CI: don't need to build deps from source (#305)"

scpeters:
This reverts commit 04854e7869a9da90208ea5c4999db3c220bc43c0. I failed to update the dependencies script, so a lot of things were broken by #305. This just reverts it for now, and we can take a more careful approach.


chapulina:
I think this can be re-reverted after #317

PR #318, opened 2020-10-09
Fix XML generation for the ghprb plugin in DSL

j-rivero:
The configuration for the Github Pull Request plugin needs to be updated to avoid log problems coming from the fact that cron is not the right XML but spec. The other small change is to remove the duplicate `useGitHubHooks`.


scpeters:
can we test this on build.osrfoundation.org?

j-rivero:
run ignition DSL with the branch [Build Status](https://build.osrfoundation.org/view/All/job/_dsl_ignition/1043/) . Let's see today if there is any problem.

scpeters (approved):

PR #322, opened 2020-10-14
Whitelist main branch for github integration

j-rivero:


chapulina (approved):
wow, I didn't know the testing repository was being used for anything :sweat_smile:

j-rivero:
> wow, I didn't know the testing repository was being used for anything sweat_smile

it is! in fact this bug ended up in a nice debugging on Jenkins github integration.

PR #324, opened 2020-10-20
Limit ABI and debbuild jobs to max of 200 builds

j-rivero:
Found Jenkins consuming extensive CPU and one of the reasons could be the more 9k builds that are kept for debbuild or abi jobs. This change should keep a max of 200.


scpeters:
have you tested this branch?

scpeters (approved):

j-rivero:
> have you tested this branch?

Sorry did not mention: I run the local generation and grep to check that changes are propagated.

scpeters:
> > have you tested this branch?
> 
> Sorry did not mention: I run the local generation and grep to check that changes are propagated.

sounds good; go ahead and merge at your leisure

Issue #331, opened 2020-10-26
Edifice bump: ign-rendering5 and dependencies


chapulina:
This is to keep track of everything that needs to be updated to use `ign-rendering5` on Edifice.

All these libraries will need to be bumped:

* `ign-rendering5`
* `ign-gazebo5`
* `ign-sensors5`
* `ign-gui5`
* `ign-launch4`

- [x] Source code of all downstream libraries
    * https://github.com/ignitionrobotics/ign-gazebo/pull/430
    * https://github.com/ignitionrobotics/ign-gui/pull/141
    * https://github.com/ignitionrobotics/ign-launch/pull/69
    * https://github.com/ignitionrobotics/ign-sensors/pull/55
    * https://github.com/ignitionrobotics/ign-edifice/pull/1
- [x] Release repository of all downstream libraries
    * https://github.com/ignition-release/ign-gazebo5-release/pull/1
    * https://github.com/ignition-release/ign-gui5-release/pull/1
    * https://github.com/ignition-release/ign-launch4-release/pull/1
    * https://github.com/ignition-release/ign-sensors5-release/pull/1
    * https://github.com/ignition-release/ign-edifice-release/pull/1
- [x] gazebodistro
    * https://github.com/ignition-tooling/gazebodistro/pull/13
- [x] gzdev to use nightlies
    * https://github.com/ignition-tooling/gzdev/pull/21
- [x] Jenkins to build from source
    * https://github.com/ignition-tooling/release-tools/pull/332
- [x] homebrew-simulation
    * https://github.com/osrf/homebrew-simulation/pull/1191
    * https://github.com/osrf/homebrew-simulation/pull/1192
    * https://github.com/osrf/homebrew-simulation/pull/1193
    * https://github.com/osrf/homebrew-simulation/pull/1194
    * https://github.com/osrf/homebrew-simulation/pull/1195
- [x] docs
    * https://github.com/ignitionrobotics/docs/pull/112

(I'm writing a script to help with this in the future)

chapulina:
The script that I used to open all these PRs is here: https://github.com/ignition-tooling/release-tools/pull/333

chapulina:
Merge order:

1. Homebrew / release-tools / gazebodistro / gzdev
1. source code
1. release repositories
1. docs

We shouldn't wait too long between the steps though, the moment some things are merged we're left in a bad state. Ideally these would all go in in the same day.

chapulina:
[...]

PR #333, opened 2020-10-29
Script to bump a library's version in the upcoming collection

chapulina:
I used iterations of this script to open all the pull requests connected to https://github.com/ignition-tooling/release-tools/issues/331, except for the `gzdev` one.

The script still needs a lot of work, but it's a great improvement over doing all of that by hand. Some TODO items that stand out:

* The script relies on most of the PRs not being merged. After https://github.com/ignition-tooling/gazebodistro/pull/13 was merged, I had to pin to the previous commit.
* Running the script twice shouldn't change the output.
* The `release-tools` part still has a manual step
* Updating `gzdev` is not included yet
* ~~Updating `ign-edifice`, `ign-edifice-release` and `ign-edifice` on homebrew~~ ff4ac9ce2f6dac66327c9df4ba938410ee88b42e
* The coloring could be more consistent to make it easier to navigate the logs

We'll have a chance to test this script again soon when we bump `sdformat` and dependencies for Edifice. 

Also started a README so we can start documenting what's in this repository to improve discoverability.

### DIFF
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -303,11 +303,6 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
 
   LIB=${LIBRARIES[$i]}
 
-  if ! [ "$LIB" = "ign-fortress" ]; then
-    continue
-  fi
-
-
   LIB_=${LIB//-/_} # For fuel_tools
   VER=${VERSIONS[$i]}
   PREV_VER="$((${VER}-1))"
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -1,0 +1,436 @@
+# Copyright (C) 2020 Open Source Robotics Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# The script will bump the major version of a library for a given collection
+#
+# Use:
+# $ ./bump_dependency.bash <collection> <library> <version>
+
+# TODO: Update gzdev to use nightlies
+
+DEFAULT="\e[39m"
+DEFAULT_BG="\e[49m"
+
+GREY="\e[90m"
+GREEN="\e[32m"
+RED="\e[31m"
+WHITE_BG="\e[107m"
+BLUE_BG="\e[44m"
+GREEN_BG="\e[42m"
+
+COLLECTION=${1}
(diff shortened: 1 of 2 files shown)


release-repo-scripts/bump_dependency.bash
```diff
@@ -0,0 +1,455 @@
...
+  ##################
+  # source code
+  ##################
+
+  echo -e "${GREEN}${LIB}: source code${DEFAULT}"
+
+  cd ${TEMP_DIR}
+  if [ ! -d "${LIB}" ]; then
+    echo -e "${GREEN}Cloning ${LIB}${DEFAULT}"
+    git clone https://github.com/ignitionrobotics/${LIB}
```
azeey, line 226: This always clones from `https://github.com/ignitionrobotics/${LIB}`, which won't work for sdformat. Can we maybe use `vcstool`?

chapulina: > Can we maybe use vcstool?

I'm open to that, the thing is that the script isn't following a yaml file. There may be a better way of doing this, but the libraries are being added to the `LIBRARIES` array recursively. Let me see if I can do something quick to support SDF.

Another assumption that the script makes is that the unreleased version is `main`, but for SDF that's `master`.

And I noticed that there's no `sdformat11-release` branch, which means we aren't making nightlies... Lots to fix, let me address some of this.

chapulina: There should be better support for SDFormat now: c0467fcd90bbb57cec33c31b00abc0677cae75e7

I didn't check it too closely though, there may be more gotchas in there.

release-repo-scripts/bump_dependency.bash
```diff
@@ -0,0 +1,471 @@
+# Copyright (C) 2020 Open Source Robotics Foundation
```
mjcarroll, line 1:
```suggestion
#!/usr/bin/env bash
# Copyright (C) 2020 Open Source Robotics Foundation
```

chapulina:
This script was used to address #331, #360, #361 and #362. It still requires some manual cleanup here and there, but in general I think it's a large speed up over doing all this manually. There's a lot of room for improvement, for example, using the GitHub cli to open PRs with the correct text would help even more. I wish someone else had had the chance to use it until the end, but unfortunately time didn't permit :confused: 

My suggestion is to get this in for now and work on improvements on top of it as we go. Let me know what you'd prefer, @j-rivero .

j-rivero:
> confused
> 
> My suggestion is to get this in for now and work on improvements on top of it as we go. Let me know what you'd prefer, @j-rivero .

I agree with merging current limited support and keep improving it as long as the features enter into plans or a bit of free time to do them appear.

j-rivero (approved):

chapulina:
> I agree with merging current limited support 

Thanks, Jose! The script is in a intermediate state right now while I transition to the latest format:

* Open separate `gazebodistro` PRs for each library
* Open PRs automatically using the `gh` CLI
* Enable nightlies for the bumped versions

I hope to wrap it up by the end of the day and trigger all PRs for #428 , and then we can merge it and revisit it for Ign-G.

chapulina:
Ok, I triggered all PRs in #428 using the latest version of the script. I think this is good as a first revision.

chapulina:
Merging this for now, we can revisit it when we need it in 6 months :grimacing:

PR #337, opened 2020-11-02
👩‍🌾 Disable Gazebo7 CI on macOS

chapulina:
Related to https://github.com/osrf/homebrew-simulation/issues/1110

I think it would be good to also disable CI for PRs targeting `gazebo7`, but I'm not sure where to do that.

### DIFF
--- a/jenkins-scripts/dsl/gazebo.dsl
+++ b/jenkins-scripts/dsl/gazebo.dsl
@@ -549,6 +549,9 @@ all_branches.each { branch ->
 
   gazebo_brew_ci_job.with
   {
+      if (("${branch}" == "gazebo7"))
+        disabled()
+
       label osx_label
 
       triggers {


scpeters (approved):

scpeters:
> I think it would be good to also disable CI for PRs targeting `gazebo7`, but I'm not sure where to do that.

The ABI any to any jobs use a whitelist of branches that are allowed as the target and exclude `main` from that whitelist:

* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/ignition.dsl#L230-L232
* https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy#L17

I suppose that could be done for gazebo's homebrew `pr_any` jobs, but I think that would only allow CI builds for PR's targeting a branch in that whitelist, so CI wouldn't fully run on PR's targeting another PR for example

chapulina:
Ok, I think we can punt on disabling PR CI for now.

PR #347, opened 2020-11-20
vcpkg snapshot support: check and upgrade

j-rivero:
This should close https://github.com/ignition-tooling/release-tools/issues/340 (also help in https://github.com/ignitionrobotics/ign-common/issues/122). The idea is to define a which snapshot of vcpkg we are using (it matches with its github releases), use it statically until we find a reason to upgrade. To help with the upgrade, this PR creates two jobs:
 *  ` _vcpkg_testing_snapshot`: uses a special label `wintesting` to run the upgrade of the snapshot together with trying to build `ign-gazebo` (which should exercise most of the ignition software given its dependency chain).
 * `_vcpkg_update_snapshot`: the job has a parameter to inject a node name to run the upgrade of vcpkg (beware, takes long time). Instead of making it to run in all nodes in parallel I decided to go with the selected node option (b30102f ) to keep the buildfarm going while we are upgrading.

The job also includes some missing dependencies in ign-gazebo, 12c7648

Testing:
 * DSL generation: [Build Status](https://build.osrfoundation.org/job/_dsl_core/740/)
 * Testing job [upgrading vcpkg snapshot](https://build.osrfoundation.org/job/_vcpkg_testing_snapshot/50/consoleFull#console-section-4)
 * Testing job [trying to build ign-gazebo  ](https://build.osrfoundation.org/job/_vcpkg_testing_snapshot/53/consoleFull#console-section-8)
 * Testing [ffmpeg installed job](https://build.osrfoundation.org/job/_vcpkg_testing_snapshot/54/console). igniion-gazebo is failing because a protobuf problem but igntion-common was able to be built with av support. 

Note: we need to upgrade nodes before merging. Otherwise they will fail in the check of the right snapshot being used.

### DIFF
--- a/jenkins-scripts/dsl/core.dsl
+++ b/jenkins-scripts/dsl/core.dsl
@@ -9,10 +9,16 @@ def update_vcpkg_snapshot_job = job("_vcpkg_update_snapshot")
 OSRFWinBase.create(update_vcpkg_snapshot_job)
 update_vcpkg_snapshot_job.with
 {
+    parameters {
+        nodeParam('TARGET_NODE') {
+            description('Node to be updated')
+        }
+    }
+
     steps
     {
       batchFile("""\
-            call "./scripts/jenkins-scripts/vcpkg-bootstrap.bat
+            call "%WORKSPACE%/scripts/jenkins-scripts/vcpkg-bootstrap.bat
             """.stripIndent())
     }
 }
--- a/jenkins-scripts/dsl/core.dsl
+++ b/jenkins-scripts/dsl/core.dsl
@@ -1,0 +1,41 @@
+import _configs_.*
+import javaposse.jobdsl.dsl.Job
+
+def NO_TESTING = false
+def NO_BRANCHES = []
+def NO_GITHUB_PR_INTEGRATION = false
+
+def update_vcpkg_snapshot_job = job("_vcpkg_update_snapshot")
+OSRFWinBase.create(update_vcpkg_snapshot_job)
+update_vcpkg_snapshot_job.with
+{
+    steps
+    {
+      batchFile("""\
+            call "./scripts/jenkins-scripts/vcpkg-bootstrap.bat
+            """.stripIndent())
+    }
+}
+
+
+def ignition_testing_software = 'gazebo'
+def testing_vcpkg_job = job("_vcpkg_testing_snapshot")
+OSRFWinCompilationAnyGitHub.create(testing_vcpkg_job,
+                                  "ignitionrobotics/ign-${ignition_testing_software}",
+                                  NO_TESTING, NO_BRANCHES, NO_GITHUB_PR_INTEGRATION)
+testing_vcpkg_job.with
+{
+    parameters {
+        stringParam('VCPKG_SNAPSHOT', '','vcpkg tag/release to test')
+    }
+
+    steps
+    {
+      label "win_testing"
+
+      batchFile("""\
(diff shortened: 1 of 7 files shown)


jenkins-scripts/ign_gazebo-default-devel-windows-amd64.bat
```diff
@@ -5,7 +5,7 @@ set VCS_DIRECTORY=ign-gazebo
 set PLATFORM_TO_BUILD=x86_amd64
 set IGN_CLEAN_WORKSPACE=true
 :: dlfcn
-set DEPEN_PKGS="dlfcn-win32 cuda curl jsoncpp freeimage ogre gts glib fcl eigen3 ccd assimp libyaml libzip gflags protobuf tinyxml2"
+set DEPEN_PKGS="dlfcn-win32 cuda cppzmq curl jsoncpp ffmpeg freeimage ogre qt5 qwt gts glib fcl eigen3 ccd assimp libyaml libzip gflags protobuf tinyxml2 zeromq"
```
chapulina, line 8: Do all transient dependencies need to be listed here? Wouldn't other Ignition libraries bring these?

jenkins-scripts/dsl/core.dsl
```diff
@@ -0,0 +1,47 @@
+import _configs_.*
+import javaposse.jobdsl.dsl.Job
+
+def NO_TESTING = false
+def NO_BRANCHES = []
+def NO_GITHUB_PR_INTEGRATION = false
+
+def update_vcpkg_snapshot_job = job("_vcpkg_update_snapshot")
```
chapulina, line 8: Are we going to keep this name for the job? Will this substitute our existing Windows jobs?

chapulina (commented):
>  igntion-common was able to be built with av support.

Nice! Is there a way we can trigger just an `ign-common` build?

jenkins-scripts/ign_gazebo-default-devel-windows-amd64.bat
j-rivero: We don't currently have a way of importing vcpkg dependencies from software dependencies nicely. So yes, we would need to list here all transition dependencies here by now or implement a way of doing it in these scripts.

jenkins-scripts/dsl/core.dsl
j-rivero: The job is designed only to manage the updates of the vcpkg snapshot, it is doing something we did not have before so it can not replace any of the existing scripts. Since it is an internal job to control parts of infrastructure I'm using the `_` prefix.

chapulina: Ahhh gotcha, I completely misunderstood :smile:

chapulina (approved):
I don't understand 70% of it, but I didn't see anything fishy :fish:

scpeters (approved):
sorry I didn't finish reviewing this sooner; it looks pretty good to me

jenkins-scripts/ign_common-default-devel-windows-amd64.bat
```diff
@@ -4,4 +4,6 @@ set VCS_DIRECTORY=ign-common
 set PLATFORM_TO_BUILD=x86_amd64
 set IGN_CLEAN_WORKSPACE=true
 
+set DEPEN_PKGS="ffmpeg"
```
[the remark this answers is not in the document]
nuclearsandwich: Attempted to address this comment in #355

chapulina:
> we need to upgrade nodes before merging

Argh sorry about that, was too excited to get this in!

PR #358, opened 2020-12-10
README: more release instructions

scpeters:
* Watch debbuilder jobs to ensure they don't fail.
* Add "build bottle" comment to osrf/homebrew-simulation PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,19 @@ Gazebo Classic and Ignition software.
     1. If the dry run succeeds, run the same command again, now without `--dry-run`.
 1. Check that:
     * Several `-debbuilder` jobs have been queued in https://build.osrfoundation.org/ and watch those jobs to see if any of them fail.
+      While it would be easier to track the status of these jobs if there was a [dashboard](https://github.com/ignition-tooling/release-tools/issues/295),
+      you can watch the page for a specific debbuild, such as https://build.osrfoundation.org/job/ign-gui3-debbuilder/.
+        1. If you observe a failure of an Ignition Dome package, check the list of supported
+           architectures in [issue #297](https://github.com/ignition-tooling/release-tools/issues/297);
+           for example Bionic/armhf and Focal/armhf are not supported for Dome.
+           To disable builds for a specific architecture, add a file to the `-release` repository that starts with `.releasepy_NO_ARCH_`
+           and append the name of the architecture to be excluded. The file can be added to the
+           root of the `-release` repository (like [.releasepy_NO_ARCH_ARM64 in ign-gazebo2-release](https://github.com/ignition-release/ign-gazebo2-release/blob/master/.releasepy_NO_ARCH_ARM64))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,7 @@ Gazebo Classic and Ignition software.
         1. Classic release: `<path to release-tools>/release.py -r 1 gazebo11 11.2.0 secrettoken --dry-run`
     1. If the dry run succeeds, run the same command again, now without `--dry-run`.
 1. Check that:
-    * Several `-debbuilder` jobs have been queued in https://build.osrfoundation.org/
+    * Several `-debbuilder` jobs have been queued in https://build.osrfoundation.org/ and watch those jobs to see if any of them fail.
```
adlarkin, line 64: Should we mention things users can do if any of the jobs fail?

README.md
```diff
@@ -61,6 +61,7 @@ Gazebo Classic and Ignition software.
...
     * A pull request was opened to https://github.com/osrf/homebrew-simulation
         1. This pull request may take a minute or two to open.
+        1. Once it is open, make a comment containing the text "build bottle"
```
adlarkin, line 67: Should we add an example link? Something like this: https://github.com/osrf/homebrew-simulation/pull/1225#issuecomment-742870169

Also, should we mention what you should check for/expect after commenting "build bottle"?

adlarkin (commented):
A few suggestions to give users some more clarity when following these instructions.

JShep1 (commented):
Just a general comment as I'm currently doing this and it's a question I have.  When should the PR that's opened on `homebrew-simulation` be merged?

README.md
scpeters: I've added a link to the osrf/homebrew-simulation README in 9cc73a90070cebaec8ec82c3d4db825ce6a11761, since I think it will be easier to keep that info over there.

> should we mention what you should check for/expect after commenting "build bottle"?

can you look at the linked readme and see if that answers your question? if not, let's follow up in that other repo

scpeters:
> Just a general comment as I'm currently doing this and it's a question I have. When should the PR that's opened on `homebrew-simulation` be merged?

I just added a reference to the README for osrf/homebrew-simulation in 9cc73a9. That's where this information should be located. I expect there's room for improvement. Can you open an issue / pull request on the Readme that repo if the text doesn't fully answer your question?

README.md
scpeters: I'm thinking about this and trying to figure out what those things are

scpeters: I've added some additional steps to take in 6bc264e. Please take a look

adlarkin: This looks good - the only thing I would consider adding is a link to http://packages.osrfoundation.org/gazebo/ - we can mention that users can see previous releases here if they're troubleshooting failed builds and want to see if a particular build has failed in the past

scpeters: how does ad48124 look?

scpeters:
any chance for an approval here? I'll merge at the end of the day unless I any further requests for changes

README.md
adlarkin: Looks great!

adlarkin (approved):
Thanks for writing this up, I think it looks great!

PR #364, opened 2020-12-16
Add sha1 default branch in -pr- and fix XML configuration for crontab polling

j-rivero:
Should workaround https://github.com/ignition-tooling/release-tools/issues/242 in two ways:

* If the [my comment about](https://github.com/ignition-tooling/release-tools/issues/242#issuecomment-746099647) `<spec>` and `<cron>` is correct and there is a miss-configuration in the XML that leads to polling operations, I've verified that with the definition of `<cron>` as empty there is no default value anymore coming to crontab when opening the Jenkins GUI manually.
* I'm adding a default `sha1` value here following instructions of the Jenkins plugin [that explains](https://github.com/jenkinsci/ghprb-plugin):

> If you want to manually build the job, in the job setting check This build is parameterized and add string parameter named sha1 with a default value of master.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy
+++ b/jenkins-scripts/dsl/_configs_/GenericAnyJobGitHub.groovy
@@ -30,7 +30,7 @@ class GenericAnyJobGitHub
     {
       parameters
       {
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
Looks reasonable to me.

j-rivero:
> Looks reasonable to me.

thanks Jacob!

PR #365, opened 2020-12-16
Use colcon + vcpkg for sdformat10 and newer

j-rivero:
We still have some scripts that relies on the old behaviour of using tarballs + configure.bat files. Here I'm enabling colcon + vcpkg to sdformat10 and above. This should fix the problem with tinyxml2.

Tested here: [Build Status](https://build.osrfoundation.org/job/sdformat-ci-sdformat10-windows7-amd64/21/)

The failing tests can come from the fact of compiling ign-tools support (in gazebodistro file) that was previously ignored.

//cc @scpeters @azeey

### DIFF
--- a/jenkins-scripts/sdformat-default-devel-windows-amd64.bat
+++ b/jenkins-scripts/sdformat-default-devel-windows-amd64.bat
@@ -15,7 +15,7 @@ if %SDFORMAT_MAJOR_VERSION% GEQ 10 (
   set COLCON_PACKAGE=sdformat
   set COLCON_AUTO_MAJOR_VERSION=true
   call "%SCRIPT_DIR%\lib\colcon-default-devel-windows.bat"
-else if %SDFORMAT_MAJOR_VERSION% GEQ 6 (
+) else if %SDFORMAT_MAJOR_VERSION% GEQ 6 (
   call "%SCRIPT_DIR%/lib/generic-default-devel-windows.bat"
 ) else (
   call "%SCRIPT_DIR%/lib/sdformat-base-windows.bat"
--- a/jenkins-scripts/sdformat-default-devel-windows-amd64.bat
+++ b/jenkins-scripts/sdformat-default-devel-windows-amd64.bat
@@ -9,7 +9,13 @@ set IGN_CLEAN_WORKSPACE=true
 set DEPEN_PKGS="libxml2 tinyxml2"
 
 for /f %%i in ('python "%SCRIPT_DIR%\tools\detect_cmake_major_version.py" "%WORKSPACE%\%VCS_DIRECTORY%\CMakeLists.txt"') do set SDFORMAT_MAJOR_VERSION=%%i
-if %SDFORMAT_MAJOR_VERSION% GEQ 6 (
+
+if %SDFORMAT_MAJOR_VERSION% GEQ 10 (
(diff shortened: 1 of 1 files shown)


azeey:
LGTM! Do we need to disable the ign tests for windows?

j-rivero:
> Do we need to disable the ign tests for windows?

It would  be nice to fix them. Meanwhile, could you please open an issue in sdformat repository about it?

jenkins-scripts/sdformat-default-devel-windows-amd64.bat
```diff
@@ -9,7 +9,13 @@ set IGN_CLEAN_WORKSPACE=true
 set DEPEN_PKGS="libxml2 tinyxml2"
 
 for /f %%i in ('python "%SCRIPT_DIR%\tools\detect_cmake_major_version.py" "%WORKSPACE%\%VCS_DIRECTORY%\CMakeLists.txt"') do set SDFORMAT_MAJOR_VERSION=%%i
-if %SDFORMAT_MAJOR_VERSION% GEQ 6 (
+
+if %SDFORMAT_MAJOR_VERSION% GEQ 10 (
+  :: This needs to be migrated to DSL to get multi-major versions correctly
```
scpeters, line 14: I'm not sure what this comment means. I think we only check which distro to use in the dsl; I believe we always check the major version in scripts. we certainly know the major version for the release-branch jobs, but not the `ci-pr_any`

scpeters:
is the problem that `generic-default-devel-windows.bat` doesn't install any vcpkg packages?

jenkins-scripts/sdformat-default-devel-windows-amd64.bat
j-rivero: I agree, not sure where it comes from, removed.

j-rivero:
> is the problem that `generic-default-devel-windows.bat` doesn't install any vcpkg packages?

that is correct.

scpeters:
> > Do we need to disable the ign tests for windows?
> 
> It would be nice to fix them. Meanwhile, could you please open an issue in sdformat repository about it?

I think we should just disable them. I don't think I have seen a single package with a working `ign_TEST` on Windows

j-rivero:
> I think we should just disable them. I don't think I have seen a single package with a working `ign_TEST` on Windows

ook! Should I use `ign-cmake2` or `ign-utils` for `IGN_UTILS_TEST_DISABLED_ON_WIN32` macro in sdformat10?

scpeters:
> > I think we should just disable them. I don't think I have seen a single package with a working `ign_TEST` on Windows
> 
> ook! Should I use `ign-cmake2` or `ign-utils` for `IGN_UTILS_TEST_DISABLED_ON_WIN32` macro in sdformat10?

we aren't yet using ign-utils in sdformat, so I would use ign-cmake2 for now

cc @azeey

j-rivero:
Test using the branch that disables ign test on Windows [Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-windows7-amd64/1802/)

scpeters:
[Build Status](https://build.osrfoundation.org/job/sdformat-ci-sdformat10-windows7-amd64/23/) https://build.osrfoundation.org/job/sdformat-ci-sdformat10-windows7-amd64/23/

scpeters (approved):

PR #371, opened 2021-01-05
WIP: Fix homebrew_pr_bottle_hash_updater

scpeters:
Our homebrew bottle workflow has been broken since the https://build.osrfoundation.org/job/generic-release-homebrew_pr_bottle_hash_updater/ job started failing recently. This job runs at the end after bottles have been built and pushes a commit to the open pull request with the new bottle `sha256` values. Something has been failing when trying to update ruby gems when calling `brew audit`. I copied the `restore_brew()` bash function from [_homebrew_cleanup.bash](https://github.com/ignition-tooling/release-tools/blob/32a0ef7c573dcfec9cc4c2e9f450a7979df9a21b/jenkins-scripts/lib/_homebrew_cleanup.bash#L4) (used on macOS), but that doesn't seem to help, as it complains that `libz` is missing from the base system. I tried a naive approach to install `libz-dev` in 8bd76db but that doesn't work either.

Help me Jose Luis Rivero, you're my only hope.

### DIFF
--- a/jenkins-scripts/lib/_homebrew_github_setup.bash
+++ b/jenkins-scripts/lib/_homebrew_github_setup.bash
@@ -6,6 +6,8 @@
 
 restore_brew()
 {
+    sudo apt update
+    sudo apt install libz-dev
     rm -fr /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/bundle/ruby
     ${BREW} update-reset
     ${BREW} vendor-install ruby
--- a/jenkins-scripts/lib/_homebrew_github_setup.bash
+++ b/jenkins-scripts/lib/_homebrew_github_setup.bash
@@ -4,6 +4,13 @@
 # Return:
 # -> TAP_PREFIX
 
+restore_brew()
+{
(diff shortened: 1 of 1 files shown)


j-rivero:
Although my short-term plan is to move the job out of master, I see no reason to delay the fix for this problem so I've installed the `libz-dev` package in the server. Seems to be working now: [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_pr_bottle_hash_updater/1149/)

j-rivero (approved):

scpeters:
> Although my short-term plan is to move the job out of master, I see no reason to delay the fix for this problem so I've installed the `libz-dev` package in the server. Seems to be working now: [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_pr_bottle_hash_updater/1149/)

if all we needed was to install `libz-dev` in master, I'm not sure we need this PR, since the `sudo apt` lines were failing when I last tried them

* https://build.osrfoundation.org/job/generic-release-homebrew_pr_bottle_hash_updater/1143/console

I'll just close this for now

PR #384, opened 2021-01-27
Use VCPKG_MAX_CONCURRENCY to limit compilation threads on Windows

j-rivero:
Needed to avoid internal compiler errors when compiling large packages such as ogre or qt.


j-rivero:
going to merge it to put the new node online

PR #393, opened 2021-02-05
Simplify release.py for ignition releases

j-rivero:
The PR removes death code using argument  `--ignition-repo` (there is no release information outside of `ignition-release` if I'm not wrong). It also enable the auto calculation of `package-alias` for anything package starting with `ign-` unless the option `---no-ignition-auto` is enabled (to cover hidden corner cases).

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,9 +56,10 @@ Gazebo Classic and Ignition software.
     1. `cd <path to source code>`
     1. Make sure you're in the branch to be released, after the pull request bumping the version has been merged.
     1. Dry-run `release.py` with the appropriate arguments. Some examples:
-        1. New Ignition minor release: `<path to release-tools>/release.py -r 1 ign-physics3 3.1.0 secrettoken --dry-run`
+        1. New Ignition minor release and Gazebo classic release:
+          1. `<path to release-tools>/release.py -r 1 ign-physics3 3.1.0 secrettoken --dry-run`
+          1. `<path to release-tools>/release.py -r 1 gazebo11 11.2.0 secrettoken --dry-run`
         1. Pre-release: `<path to release-tools>/release.py -r 1 ign-physics3 3.0.0~pre1 --upload-to-repo=prerelease secrettoken --dry-run`
-        1. Classic release: `<path to release-tools>/release.py -r 1 gazebo11 11.2.0 secrettoken --dry-run`
     1. If the dry run succeeds, run the same command again, now without `--dry-run`.
 1. Check that:
     * Several `-debbuilder` jobs have been queued in https://build.osrfoundation.org/ and watch those jobs to see if any of them fail.
--- a/README.md
+++ b/README.md
@@ -56,8 +56,8 @@ Gazebo Classic and Ignition software.
     1. `cd <path to source code>`
     1. Make sure you're in the branch to be released, after the pull request bumping the version has been merged.
     1. Dry-run `release.py` with the appropriate arguments. Some examples:
-        1. New Ignition minor release: `<path to release-tools>/release.py -r 1 ign-physics3 3.1.0 --ignition-repo -a ignition-physics3 secrettoken --dry-run`
-        1. Pre-release: `<path to release-tools>/release.py -r 1 ign-physics3 3.0.0~pre1 --ignition-repo -a ignition-physics3 --upload-to-repo=prerelease secrettoken --dry-run`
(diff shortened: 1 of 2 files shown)


chapulina:
Can we update the examples on the README to use the new commands?

release.py
```diff
@@ -129,21 +127,23 @@ def parse_args(argv):
...
                         help='extra OSRF repository to use in the build')
     parser.add_argument('--nightly-src-branch', dest='nightly_branch', default="main",
                         help='branch in the source code repository to build the nightly from')
     args = parser.parse_args()
+
+    IGN_AUTO_DETECTED = args.package.startswith('ign-') if not args.no_ignition_auto else False
+    print(IGN_AUTO_DETECTED)
     if not args.package_alias:
-        args.package_alias = args.package
+        args.package_alias = args.package.replace('ign-', 'ignition-') if IGN_AUTO_DETECTED else args.package
```
scpeters, line 143: nit: it's just my opinion but putting the logic all on one line makes it a bit harder to follow (this line and the one above)

release.py
```diff
@@ -451,8 +451,7 @@ def generate_upload_tarball(args):
         tarball_name = "mentor2"
 
     # For ignition, we use the alias without version numbers as package name
-    if IGN_REPO:
-        tarball_name = re.sub(r'[0-9]+$','', args.package_alias)
+    tarball_name = re.sub(r'[0-9]+$', '', args.package_alias)
```
scpeters, line 454: I was trying to figure out why this line was limited to `IGN_REPO`; I'm guessing because of https://github.com/osrf/mentor2 just a few lines above? We can probably remove those lines at this point since that project is long over

at any rate, I think this line is safe

scpeters (changes requested):
> Can we update the examples on the README to use the new commands?

:+1: we should update the README examples

release.py
j-rivero: let's try in a different way: d4d3375

release.py
j-rivero: I agree. I removed the mentor2/haptix references. a95ec0e

j-rivero:
> Can we update the examples on the README to use the new commands?


Ouch missed them: 10a2e15

README.md
```diff
@@ -56,8 +56,8 @@ Gazebo Classic and Ignition software.
     1. `cd <path to source code>`
     1. Make sure you're in the branch to be released, after the pull request bumping the version has been merged.
     1. Dry-run `release.py` with the appropriate arguments. Some examples:
-        1. New Ignition minor release: `<path to release-tools>/release.py -r 1 ign-physics3 3.1.0 --ignition-repo -a ignition-physics3 secrettoken --dry-run`
-        1. Pre-release: `<path to release-tools>/release.py -r 1 ign-physics3 3.0.0~pre1 --ignition-repo -a ignition-physics3 --upload-to-repo=prerelease secrettoken --dry-run`
+        1. New Ignition minor release: `<path to release-tools>/release.py -r 1 ign-physics3 3.1.0 secrettoken --dry-run`
+        1. Pre-release: `<path to release-tools>/release.py -r 1 ign-physics3 3.0.0~pre1 --upload-to-repo=prerelease secrettoken --dry-run`
         1. Classic release: `<path to release-tools>/release.py -r 1 gazebo11 11.2.0 secrettoken --dry-run`
```
scpeters, line 61: nit: I think the 1st and 3rd sub-items are now redundant and could be combined ("New Ignition minor release" and "Classic release")

release.py
```diff
@@ -129,21 +127,24 @@ def parse_args(argv):
...
     parser.add_argument('--extra-osrf-repo', dest='extra_repo', default="",
                         help='extra OSRF repository to use in the build')
     parser.add_argument('--nightly-src-branch', dest='nightly_branch', default="main",
                         help='branch in the source code repository to build the nightly from')
     args = parser.parse_args()
-    if not args.package_alias:
-        args.package_alias = args.package
+
+    args.package_alias = args.package
+    # If ignition auto is enabled, replace ign- at the begging of the string
```
scpeters, line 141: nit: "replace ign- at the begging" --> "replace ign- with ignition- at the beginning"

scpeters (approved):
thanks, looks good!

release.py
j-rivero: 38ac800

README.md
j-rivero: 200d153

PR #394, opened 2021-02-05
Implement --only-bump-revision-linux in release.py

j-rivero:
The PR adds the `--only-bump-revision-linux` that allows to call Jenkins on revision bumps (those only changing build metadata no source code from the application) without uploading tarballs.

The current code is a bit messy and to get the tarball name that was finally uploaded in the first time it is needed to generate locally the tarball, although it is not finally uploaded. 

recommendation: enable w=1 to see differences in code https://github.com/ignition-tooling/release-tools/pull/394/files?w=1

### DIFF
--- a/release.py
+++ b/release.py
@@ -137,7 +137,7 @@ def parse_args(argv):
                         help='extra OSRF repository to use in the build')
     parser.add_argument('--nightly-src-branch', dest='nightly_branch', default="main",
                         help='branch in the source code repository to build the nightly from')
-    parser.add_argument('--only-bump-revision', dest='bump_revision',
+    parser.add_argument('--only-bump-revision-linux', dest='bump_rev_linux_only',
                         action='store_true', default=False,
                         help='Bump only revision number. Do not upload new tarball.')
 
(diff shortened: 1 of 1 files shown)


scpeters:
I think this will conflict with #393, which should we review first?

j-rivero:
> I think this will conflict with #393, which should we review first?

Don't have any preference. Since we have already started with #393 , let's finish it first.

chapulina:
I haven't looked at the code, but I just used this branch to trigger `ign-transport10.0.0~pre2-2` I think it worked, but I see the `tar` line, which I'm not sure should be there.

```
Running:
  tar xf /tmp/tmpfs3q49t8/orig.tar -C /tmp/tmpfs3q49t8/src
```

scpeters:
> I haven't looked at the code, but I just used this branch to trigger `ign-transport10.0.0~pre2-2` I think it worked, but I see the `tar` line, which I'm not sure should be there.
> 
> ```
> Running:
>   tar xf /tmp/tmpfs3q49t8/orig.tar -C /tmp/tmpfs3q49t8/src
> ```

did you see the following message in the console output:

~~~
INFO: bump revision is enabled. It needs to generate a local tarball although it will not upload it
~~~

j-rivero:
> > ```
> > 
> > ```
> 
> 
> did you see the following message in the console output:

not sure if we can improve the message or user information in any other way. Seems like it worked well, @chapulina ?

chapulina:
> did you see the following message in the console output:

I didn't notice it, but it may have been there.

> Seems like it worked well,

Yup, AFAICT it worked perfectly :+1:

PR #379, opened 2021-01-22
Bump in edifice: ign-common4

chapulina:
Addresses https://github.com/ignition-tooling/release-tools/issues/361

PR #397, opened 2021-02-17
Don't build dependencies from source on CI, use nightlies

chapulina:
Ubuntu CI is already installing dependencies from nightlies, so we don't need to build them from source too.

This should save lots of CI time.

If this gets in, I'll decline #379 and update #333


scpeters (approved):
there's a block that could be removed from [ign_physics-compilation.bash](https://github.com/ignition-tooling/release-tools/blob/0971fe88c5864f3f8fb22f431deedb4cc8d9218c/jenkins-scripts/docker/ign_physics-compilation.bash#L33-L37) as well, but otherwise this looks good to me

chapulina:
Test `ign-launch` build: [Build Status](https://build.osrfoundation.org/job/ignition_launch-ci-pr_any-ubuntu_auto-amd64/378/)

PR #425, opened 2021-03-31
bottle build: don't use HOMEBREW_UPDATE_TO_TAG

scpeters:
We need to use the HEAD of homebrew/brew, not a tag. See https://github.com/osrf/homebrew-simulation/pull/1395#issuecomment-811176482


j-rivero (approved):
great, thanks

Issue #295, opened 2020-10-01
Dashboard to check what releases are available to each architecture


chapulina:
As we have more libraries to release, it becomes harder to keep track of which binary releases went through for each architecture combination. It would be helpful to have a live dashboard which displays:

* Whether a library's major version is available for a given OS + arch combination
* If available, what's the latest version number

I think that would look a bit like the ROS status pages:

http://repositories.ros.org/status_page/ros_noetic_default.html

While tracking the Dome release, I created a simple [script](https://gist.github.com/chapulina/52cb0ea5ef104c4f031753336cec68ef) that gets the job partially done, but there's a lot still hardcoded in it. It also needs to be run locally, which is less convenient than a web page.

This is what the results look like:

![image](https://user-images.githubusercontent.com/5751272/94762728-56522d00-035d-11eb-9608-ff2c527437a5.png)

chapulina:
See #426

PR #426, opened 2021-03-31
Import terminal dashboard from Louise. Display versions

j-rivero:
I just imported the terminal dashboard used by Louise during the releases and modify it to display the versions available. Nothing solid since it needs a lot of parametrization and use of gazebodistro probably. Left in draft by now:

image


terminal-dashboard/table.bash
```diff
@@ -0,0 +1,74 @@
...
+  "ignition-transport10"
+  "ignition-fuel-tools5"
+  "ignition-plugin"
+  "ignition-rendering5"
+  "ignition-physics4"
+  "ignition-sensors5"
+  "ignition-gui5"
+  "ignition-gazebo5"
+  "ignition-launch4"
+  "ignition-edifice"
```
chapulina, line 19: It would be nice to support any collection here, maybe passing a command line argument. We may even be able to leverage `gazebodistro` so we don't need to hardcode the versions.

terminal-dashboard/table.bash
```diff
@@ -0,0 +1,74 @@
...
+      do
+        PADDED_ARCH=$ARCH
+        PADDED_ARCH="${PADDED_ARCH:0:8}${COLUMN:0:$((8 - ${#PADDED_ARCH}))}"
+
+        echo -n " "
+
+        if [[ $ARCH == "i386" && $VER == "focal" ]]; then
+          PKG_VERSION="disabled"
+        else
+	  PKG_VERSION=$(wget -qO- http://packages.osrfoundation.org/gazebo/${DISTRO}-stable/dists/${VER}/main/binary-${ARCH}/Packages | grep -1 "Source: ${LIB}" | sed -n 's/^Version: \(.*\)/\1/p' | uniq)
```
chapulina, line 56: Indentation starts getting a bit messy from here down

```suggestion
	    PKG_VERSION=$(wget -qO- http://packages.osrfoundation.org/gazebo/${DISTRO}-stable/dists/${VER}/main/binary-${ARCH}/Packages | grep -1 "Source: ${LIB}" | sed -n 's/^Version: \(.*\)/\1/p' | uniq)
```

chapulina (commented):
Thanks for the improvements! I'm leaving comments with feature requests, with the full understanding that we may not be able to address them anytime soon :sweat_smile: 

---

Right now, if there's any version released for an architecture, that shows up as green. For example, see `ign-math`:

image

At a quick glance, one may think that the latest release succeeded for all platforms, but in fact, if you look closely, Focal::armhf and Buster::arm64 still have 6.7.0. It would be helpful to show older versions in a different color, so we know that the latest release failed for those archs.

terminal-dashboard/table.bash
```diff
@@ -0,0 +1,74 @@
...
+
+LIBS=(
+  "ignition-cmake2"
+  "ignition-math6"
+  "ignition-utils1"
+  "ignition-tools"
+  "ignition-common4"
+  "ignition-msgs7"
+  "ignition-transport10"
+  "ignition-fuel-tools5"
```
chapulina, line 11:
```suggestion
  "ignition-fuel-tools6"
```

j-rivero:
> Thanks for the improvements! I'm leaving comments with feature requests, with the full understanding that we may not be able to address them anytime soon sweat_smile

I agree. Some of the features you are proposing are hard to get them right using bash and would probably require a bit of real programming (like getting different colors for different versions). Before we spent more time on this, we probably want to revisit #295 . My proposal would be to leave it as-it is or with some small changes and merge and focus efforts on #295.

chapulina (approved):

PR #422, opened 2021-03-30
Stable branches for Edifice

chapulina:
Before merging this, we need to:

- [x] Release stable Edifice
- [ ] Bump versions for all `main` branches
- [x] Retire sdf's `master` and use `main` instead (https://github.com/osrf/sdformat/issues/531)
- [x] Create `ign-fortress`

https://github.com/ignition-tooling/release-tools/issues/421

PR #431, opened 2021-04-07
ignition.dsl: re-enable ign-utils1-debbuilder

scpeters:
it has been disabled since #422 was merged

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -64,7 +64,7 @@ ignition_extra_debbuild = [ 'cmake3',
                             'sensors6',
                             'tools2',
                             'transport11',
-                            'utils1',
+                            'utils1', // see comment https://github.com/ignition-tooling/release-tools/pull/431#issuecomment-815099918
                             'utils2']
 // DESC: exclude ignition from generate any install testing job
 ignition_no_pkg_yet         = [  ]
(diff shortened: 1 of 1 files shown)


scpeters:
> Since we are declaring[ that 1 is a supported branch](https://github.com/ignition-tooling/release-tools/pull/431/files#diff-1269f38223f9cd477137c0a99b7e9ff46f8eeb3b09300d2b245ec8f9381172f5R45), the debbuilder should be created automatically. I think that what is happening is that we have traditionally been using non-versioned packages for versions 0 and 1. That is no longer the case, we probably need to modify the [code that transform them into unversioned](https://github.com/ignition-tooling/release-tools/pull/431/files#diff-1269f38223f9cd477137c0a99b7e9ff46f8eeb3b09300d2b245ec8f9381172f5L180-L183).

I didn't take that approach for this PR because it is more complicated and would potentially break these debbuild jobs that we still use:

* https://build.osrfoundation.org/job/ign-plugin-debbuilder/
* https://build.osrfoundation.org/job/ign-tools-debbuilder/

j-rivero:
> I didn't take that approach for this PR because it is more complicated and would potentially break these debbuild jobs that we still use:
> 
>     * https://build.osrfoundation.org/job/ign-plugin-debbuilder/
> 
>     * https://build.osrfoundation.org/job/ign-tools-debbuilder/

Ouch, you are right. Added a comment about this in 9873594

j-rivero (approved):

PR #432, opened 2021-04-09
Revert nightlies to upcoming collection (not necessarily main)

chapulina:
As of https://github.com/ignition-tooling/release-tools/issues/319 and https://github.com/ignition-tooling/release-tools/pull/334, we started generating nightlies from `main` branches, regardless of whether those branches were going to be on the upcoming collection. We tried this during Edifice's release cycle, and noticed that:

* The nightlies for versions that aren't going into the collection, such as `ign-math7`, were not useful.
* By not having nightlies for versions such as `ign-math6`, we had to make pre-releases, and downstream libraries like SDFormat had to depend on those pre-releases. Which means that they needed to mix pre-releases and nightlies, which brings lots of problems.

So going forward, the plan is to build nightlies for all the versions known to go into the upcoming collection. The nightly list will be updated as part of the process of bumping a version. 

A downsides is that we'll have to wait until those versions are bumped in the upcoming collection to know if packaging is working for them. i.e. there's no way (for now) to test `ign-math7-release` until we use `ign-math7`.


jenkins-scripts/dsl/ignition_collection.dsl
```diff
@@ -18,25 +18,25 @@ ignition_collections = [
...
+          'fuel-tools': [ debbuild: 'ign-fuel-tools6' , branch: 'ign-fuel-tools6' ],
+          'gazebo'    : [ debbuild: 'ign-gazebo5'     , branch: 'ign-gazebo5' ],
+          'gui'       : [ debbuild: 'ign-gui5'        , branch: 'ign-gui5' ],
+          'launch'    : [ debbuild: 'ign-launch4'     , branch: 'ign-launch4' ],
+          'math'      : [ debbuild: 'ign-math6'       , branch: 'ign-math6' ],
+          'msgs'      : [ debbuild: 'ign-msgs7'       , branch: 'ign-msgs7' ],
+          'physics'   : [ debbuild: 'ign-physics4'    , branch: 'ign-physics4' ],
+          'plugin'    : [ debbuild: 'ign-plugin1'     , branch: 'ign-plugin1' ],
+          'rendering' : [ debbuild: 'ign-rendering5'  , branch: 'ign-rendering5' ],
+          'sdformat'  : [ debbuild: 'sdformat11'      , branch: 'sdformat11' ],
```
j-rivero, line 35:
```suggestion
          'sdformat'  : [ debbuild: 'sdformat11'      , branch: 'sdf11' ],
```

j-rivero (changes requested):

j-rivero (approved):

PR #454, opened 2021-04-28
Boost is required to build ign-rendering.

nuclearsandwich:
This is an attempt to resolve the following error as seen on the build
machine Locutus:

```
  Could not find pkg-config information for IgnOGRE.  It was not provided by
  the find-module for the package, nor was it explicitly passed into the call
  to ign_find_package(~).  This is most likely an error in this project's use
  of ign-cmake.
Call Stack (most recent call first):
  CMakeLists.txt:66 (ign_find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
-- Could NOT find Boost (missing: Boost_INCLUDE_DIR)
```

### DIFF
--- a/jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
+++ b/jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
@@ -5,7 +5,7 @@ set VCS_DIRECTORY=ign-rendering
 set PLATFORM_TO_BUILD=x86_amd64
 set IGN_CLEAN_WORKSPACE=true
 :: ogre2 from osrf vcpkg-ports
-set DEPEN_PKGS="dlfcn-win32 cuda freeimage ogre ogre2 gts glib"
+set DEPEN_PKGS="boost dlfcn-win32 cuda freeimage ogre ogre2 gts glib"
 :: This needs to be migrated to DSL to get multi-major versions correctly
 set COLCON_PACKAGE=ignition-rendering
 set COLCON_AUTO_MAJOR_VERSION=true
--- a/jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
+++ b/jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
@@ -5,7 +5,7 @@ set VCS_DIRECTORY=ign-rendering
 set PLATFORM_TO_BUILD=x86_amd64
 set IGN_CLEAN_WORKSPACE=true
 :: ogre2 from osrf vcpkg-ports
-set DEPEN_PKGS="boost dlfcn-win32 cuda freeimage ogre ogre2 gts glib"
+set DEPEN_PKGS="boost-core dlfcn-win32 cuda freeimage ogre ogre2 gts glib"
 :: This needs to be migrated to DSL to get multi-major versions correctly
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
This was successful in resolving the error during CI:

Before [Build Status](https://build.osrfoundation.org/job/ign_rendering-pr-win/1592/)

After [Build Status](https://build.osrfoundation.org/job/ign_rendering-pr-win/1594/)

jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
```diff
@@ -5,7 +5,7 @@ set VCS_DIRECTORY=ign-rendering
 set PLATFORM_TO_BUILD=x86_amd64
 set IGN_CLEAN_WORKSPACE=true
 :: ogre2 from osrf vcpkg-ports
-set DEPEN_PKGS="dlfcn-win32 cuda freeimage ogre ogre2 gts glib"
+set DEPEN_PKGS="boost dlfcn-win32 cuda freeimage ogre ogre2 gts glib"
```
nuclearsandwich, line 8: All of boost installed (predictably) a lot. I'm not sure if there's a specific set of boost libraries that we ought to add instead.

nuclearsandwich:
This change was enough to at least resolve the issue. My interpretation of the error message suggests that boost is actually part of the Ogre dependency chain. Is that a misread? If it isn't is a more correct solution to this issue to update the vcpkg recipe for boost?

jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
j-rivero: > All of boost installed (predictably) a lot. I'm not sure if there's a specific set of boost libraries that we ought to add instead.

There are a good number of boost packages in vcpkg https://github.com/microsoft/vcpkg/tree/master/ports . We will need to investigate where the boost message came from and what is really looking for (maybe boost-core is enouhg).

j-rivero:
> This change was enough to at least resolve the issue. My interpretation of the error message suggests that boost is actually part of the Ogre dependency chain. Is that a misread? If it isn't is a more correct solution to this issue to update the vcpkg recipe for boost?

Umm, the error reported by VS is the first build is:
```
 fatal error C1090: PDB API call failed, error code '23': (0x00000005) 
```
something suggest me that this is still a signal of flaky problems on that machine. 

I wonder if the problem with not finding Boost has been there forever, ign-rendering seems not to be using it and the [ci serie has been clean](https://build.osrfoundation.org/job/ign_rendering-ci-win/) and the [boost problem was there](https://build.osrfoundation.org/job/ign_rendering-ci-win/125/consoleText).

jenkins-scripts/ign_rendering-default-devel-windows-amd64.bat
j-rivero: boost is being used by ign-rendering since there are parts of Ogre that needs it. I change the PR to use boost-core, let's see what happened: [Build Status](https://build.osrfoundation.org/job/ign_rendering-pr-win/1696/)

j-rivero (approved):

PR #456, opened 2021-05-03
Skip Optix for ign-rendering

chapulina:


chapulina:
Oh actually building `ign-cmake` from source isn't needed, we can test it with Fortress CI, which is using the latest nightlies. I noticed on this build:

https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-ubuntu_auto-amd64/1364/consoleFull

chapulina:
Test build with `ign-rendering`'s `main` branch: [Build Status](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-ubuntu_auto-amd64/1365/)

j-rivero:
> Test build with `ign-rendering`'s `main` branch

Seems like the warning is gone, only flaky test. Great.

j-rivero:
> Seems like the warning is gone, only flaky test. Great.

I assume that we don't want to restrict the `-DSkip_Optix` parameter to any of the supported branches of ign-rendering, is this correct?

chapulina:
> I assume that we don't want to restrict the -DSkip_Optix parameter to any of the supported branches of ign-rendering

Yup, I believe the issue happens with all currently supported major versions

chapulina:
The `ign-cmake` release is out, I think this is good to go

scpeters (approved):

PR #460, opened 2021-05-13
dsl: add sdformat install_bottle jobs, add ign-utils to edifice view

scpeters:
We noticed in https://github.com/osrf/homebrew-simulation/pull/1458 that there were not `install_bottle` jobs for sdformat and that the `ign-utils` install jobs weren't included in the edifice view. This remedies both of those things and also cleans up some logic related to unsupported package versions in `ignition.dsl`

## sdformat.dsl

* [Build Status](https://build.osrfoundation.org/job/_dsl_sdformat/1012/) https://build.osrfoundation.org/job/_dsl_sdformat/1012/

~~~
Added items:
    GeneratedJob{name='sdformat4-install_bottle-homebrew-amd64'}
    GeneratedJob{name='sdformat6-install_bottle-homebrew-amd64'}
    GeneratedJob{name='sdformat9-install_bottle-homebrew-amd64'}
    GeneratedJob{name='sdformat10-install_bottle-homebrew-amd64'}
    GeneratedJob{name='sdformat11-install_bottle-homebrew-amd64'}
~~~

* [Build Status](https://build.osrfoundation.org/job/sdformat11-install_bottle-homebrew-amd64/) https://build.osrfoundation.org/job/sdformat11-install_bottle-homebrew-amd64/

## ignition_collection.dsl

* [Build Status](https://build.osrfoundation.org/job/_dsl_ignition_collection/361/) https://build.osrfoundation.org/job/_dsl_ignition_collection/361/
* https://build.osrfoundation.org/view/ign-edifice/
* https://build.osrfoundation.org/view/ign-fortress/

## ignition.dsl

* [Build Status](https://build.osrfoundation.org/job/_dsl_ignition/1156/) https://build.osrfoundation.org/job/_dsl_ignition/1156/

PR #461, opened 2021-05-13
🏁 Gazebo 7 & 10 EOL

chapulina:
Gazebo 7 & 10 have reached EOL a while back, and took `ign-math2`, `ign-transport2` and `libsdformat4` with them.

See [announcement](https://community.gazebosim.org/t/gazebo-7-10-officially-end-of-life/799) and [support table](https://github.com/ignitionrobotics/docs/blob/master/tools/versions.md)

### DIFF
--- a/jenkins-scripts/dsl/_configs_/Globals.groovy
+++ b/jenkins-scripts/dsl/_configs_/Globals.groovy
@@ -26,8 +26,7 @@ class Globals
                      'rolling'  : ['focal']]
 
    // This should be in sync with archive_library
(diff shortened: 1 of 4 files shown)


j-rivero:
I was waiting for Ubuntu Xenial/ROS Kinetic to be EOL to remove all together, I just realized that it already happened. I see no reason not to do it in different PRs.

j-rivero (approved):

chapulina:
> I was waiting

Ah sorry, I noticed this when reviewing #460, because it created `sdf4` jobs that we don't need anymore

PR #480, opened 2021-06-29
👩‍🌾 Keep last 75 builds instead of 15

chapulina:
There's enough traffic in some jobs that we often lose builds from the same day. Increasing 5x, we can reduce if it starts to take up too much space.


j-rivero (approved):
thanks!

PR #501, opened 2021-08-23
👩‍🌾 Remove reliance on bitbucket pipelines files

chapulina:
Closes https://github.com/ignition-tooling/release-tools/issues/203

We used that file to determine the test platform for each version of each library. However:

* Since migrating to GitHub we haven't really been paying attention to that file
* All files are currently pointing to [Bionic](https://github.com/search?q=org%3Aignitionrobotics+image%3A+ubuntu%3Abionic&type=code) and there are none pointing to [Focal](https://github.com/search?q=org%3Aignitionrobotics+image%3A+ubuntu%3Afocal&type=code). But Bionic is the fallback anyway in case that file doesn't exist.

So removing the support for that file, as well as the files themselves, should have no impact on our CI.

I suggest we do that to avoid further confusion, we often get questions about this file.

Whenever we need to target different platforms for each branch, we should open an issue for it and tackle it in a different way.

See this PR that removes the file and CI still passes: https://github.com/ignitionrobotics/ign-fuel-tools/pull/196

---

https://github.com/osrf/buildfarmer/issues/224

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -293,14 +293,7 @@ ignition_software.each { ign_sw ->
                 wget https://raw.githubusercontent.com/osrf/bash-yaml/master/yaml.sh -O yaml.sh
                 source yaml.sh
 
-                create_variables \${WORKSPACE}/${checkout_subdir}/bitbucket-pipelines.yml
-
                 export DISTRO=${distro}
-
-                if [[ -n \${image} ]]; then
-                  echo "Bitbucket pipeline.yml detected. Default DISTRO is ${distro}"
-                  export DISTRO=\$(echo \${image} | sed  's/ubuntu://')
-                fi
 
                 ${GLOBAL_SHELL_CMD}
 


jenkins-scripts/dsl/ignition.dsl
```diff
@@ -293,15 +293,8 @@ ignition_software.each { ign_sw ->
                 wget https://raw.githubusercontent.com/osrf/bash-yaml/master/yaml.sh -O yaml.sh
                 source yaml.sh
```
scpeters, line 294: I think these 2 lines could be deleted as well (`wget` and `source`)

scpeters:
I think the most impacted branches for this change would be the gazebo9 dependencies. I'd like to see pull requests removing the `bitbucket-pipelines.yml` files from `ign-cmake0`, `ign-math4`, `ign-common1`, and `sdf6` branches before merging this

scpeters:
> I think the most impacted branches for this change would be the gazebo9 dependencies. I'd like to see pull requests removing the `bitbucket-pipelines.yml` files from `ign-cmake0`, `ign-math4`, `ign-common1`, and `sdf6` branches before merging this

* [ ] `ign-cmake0` https://github.com/ignitionrobotics/ign-cmake/pull/181
* [ ] `ign-math4`
* [ ] `ign-common1`
* [ ] `sdf6`

chapulina:
> I think the most impacted branches for this change would be the gazebo9 dependencies

Nice catch! I didn't think of Xenial. I'll see what I can do

jenkins-scripts/dsl/ignition.dsl
chapulina: ✂️✂️✂️ 441dfbf

j-rivero:
> Nice catch! I didn't think of Xenial. I'll see what I can do

Xenial is EOL, we should not need to maintain anything on it anymore.

chapulina:
> Xenial is EOL

All our branches that relied on Xenial should have PRs open already, listed above :+1:

scpeters (approved):

PR #512, opened 2021-09-13
Do not build nightly packages with prerelease repository enabled

j-rivero:
After https://github.com/ignition-tooling/release-tools/issues/319, we don't need the prerelease repositories when building the nightlies.  This is changing https://github.com/ignitionrobotics/docs/pull/139 (Edifice).

### DIFF
--- a/jenkins-scripts/dsl/ignition_collection.dsl
+++ b/jenkins-scripts/dsl/ignition_collection.dsl
@@ -642,7 +642,7 @@ nightly_scheduler_job.with
               fi
 
               echo "releasing \${n} (from branch \${src_branch}"
-              python3 ./scripts/release.py \${dry_run_str} "\${n}" nightly "\${PASS}" --extra-osrf-repo prerelease --release-repo-branch main --nightly-src-branch \${src_branch} --upload-to-repo nightly > log || echo "MARK_AS_UNSTABLE"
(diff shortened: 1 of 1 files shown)


chapulina (approved):

chapulina:
I think this may not have been enough. The latest nightlies triggered here:

https://build.osrfoundation.org/job/ignition-fortress-nightly-scheduler/162/console

Still have the pre-release repo as part of the `OSRF_REPOS_TO_USE`, see:

https://build.osrfoundation.org/job/ign-rendering6-debbuilder/330/

j-rivero:
> I think this may not have been enough. The latest nightlies triggered here:

If I'm not wrong, the change is good but the sequence of actions to propagate the change was not:
 1. We merged this DSL change
 2. The DSL change is propagated via [`_dsl_ignition_collection` job](https://build.osrfoundation.org/job/_dsl_ignition_collection/). Yesterday [it ran at 9:39am](https://build.osrfoundation.org/job/_dsl_ignition_collection/391/)
 3. The triggered of the nightlies ran BEFORE the DSL change, so it ran with the old code, [at 9:00am](https://build.osrfoundation.org/job/ignition-fortress-nightly-scheduler/162/).

I ran a dry-run of the nightly a few minutes ago, [seems fine](https://build.osrfoundation.org/job/ignition-fortress-nightly-scheduler/ws/log/*view*/). The[ job to re-generate all configurations](https://build.osrfoundation.org/job/_dsl_generate_all_dsls/) runs daily using `@daily` in the configuration of the Jenkins job. We can modify it to run before the nightly generation. Otherwise the best way of doing this would have been to trigger the dsl job manually after merging this PR.,

chapulina:
Ahh thanks for investigating! Makes sense to me

PR #514, opened 2021-09-14
Update VCPKG_OSRF_DIR on every package installation

### LINKED ISSUE #513: windows_nuc failing to install Ogre 2.2
Example build: https://build.osrfoundation.org/job/ign_gazebo-pr-win/3059/

```
Error: while loading port `ogre22`: Port definition not found
```

j-rivero:
Before the option of `--overlay-ports` exists in vcpkg, we were using a separate variable to declare OSRF vcpkg ports. `install_osrf_vcpkg_package`. That function was updating the overlay each time.  That is not the case anymore, that function is not in use and the same `install_vcpkg_package` can handle OSRF vcpkg ports. This PR removes the old function and move the update mechanism to current one.

Tested [Build Status](https://build.osrfoundation.org/job/ign_gazebo-pr-win/3062/) 

Fixes #513

### DIFF
--- a/jenkins-scripts/lib/windows_library.bat
+++ b/jenkins-scripts/lib/windows_library.bat
@@ -242,6 +242,12 @@ goto :EOF
 set LIB_DIR=%~dp0
 call %LIB_DIR%\windows_env_vars.bat || goto :error
 call %win_lib% :check_vcpkg_snapshot || goto :error
+:: update osrf vcpkg overlay
+pushd .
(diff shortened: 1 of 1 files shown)


mjcarroll (approved):
LGTM with green test

chapulina (approved):
It fixed the ogre22 issue, now we're left with https://github.com/ignitionrobotics/ign-gazebo/issues/990

Issue #73, opened 2016-06-01
Avoid race condition when generating nightlies


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-tooling-gh-pages/#!/osrf/release-tools/issues/73)) by Steve Peters (Bitbucket: [Steven Peters](https://bitbucket.org/%7B2ccfed09-18b8-4921-8d58-15ef01092802%7D/)).**

----------------------------------------
One of the gazebo nightly builds failed recently because it was running concurrently with the uploader for the nightly sdformat deb:

* The [sdformat nightly build](http://build.osrfoundation.org/view/debbuild/job/sdformat4-debbuilder/883/) ran first

and triggered the 

* [repository uploader job](http://build.osrfoundation.org/view/debbuild/job/repository_uploader_ng/1976/) and
* [gazebo nightly build](http://build.osrfoundation.org/job/gazebo8-debbuilder/236/), which failed because the uploader job changed the apt repository (yesterday's sdformat nightly build was deleted) between the time the gazebo deb builder did `apt-get update` and `apt-get install`

chapulina:
A related issue is making sure nightlies are run in topological order for Ignition libraries. Imagine the following scenario:

1. `ign-msgs` breaks ABI on `main`
2. `ign-transport`'s nightly is run against the old `ign-msgs` nightly
3. `ign-msgs` nightly is run with new code
4. `ign-gazebo`'s nightly is run with a mix of the new `ign-msgs` and the new `ign-transport`, which is linked against the old `ign-msgs`
5. `ign-gazebo` has runtime problems because of the mixed ABI

PR #515, opened 2021-09-15
Build nightlies in topological order

chapulina:
Addresses this comment: https://github.com/ignition-tooling/release-tools/issues/73#issuecomment-820559460

I'm not sure if it would work though, Jenkins may still reorder them once they're triggered.

What do you think, @j-rivero ?


adlarkin (approved):
LGTM :+1: perhaps we can wait to see if the new builds using this in https://build.osrfoundation.org/job/ign-gazebo6-debbuilder/ are successful before merging

j-rivero:
> [](https://github.com/ignition-tooling/release-tools/issues/73#issuecomment-820559460)
> 
> I'm not sure if it would work though, Jenkins may still reorder them once they're triggered.
> 
> What do you think, @j-rivero ?

You are right about the fact that triggering the builds in this order does not guarantee that they end up being build in topological order as someone can expect. Queuing and build times will go into the mix. To implement this completely we would need to implement cross-blocks on debbuilders jobs to be sure that one don't start after all the builds in the other ones are processed.

A limitation by design is that we can not wait for a particular combination of ARCH + DISTRO, we would need to wait for all the combinations. For nightlies this is probably not a problem but for regular builds, current ARM builds could stop the farm for long time even if there are machines available at that time (i.e: ign-rendering amd64 could be blocked to start after all ign-math ARM builds are done). 

For nightlies: introducing a sleep period of a bunch of minutes between triggers can help to process them right, but again, it's mostly a workaround. 

I don't think that this PR does any harm but won't solve the root of the problem completely.

chapulina:
> I don't think that this PR does any harm but won't solve the root of the problem completely.

Thanks for the feedback. I'll leave #73 open, but merge this in the hope that we get _some_ improvement, with the full understanding that the issue is not going away completely.

PR #520, opened 2021-09-24
Allow ARM jobs to run on native arm hosts.

nuclearsandwich:
Update the label expression for ARM releases to allow them to run on
hosts with the `linux-arm64` or `linux-armhf` labels respectively in
addition to hosts with the `large-memory` label.

Example ARM host: https://build.osrfoundation.org/computer/linux-ip-172-30-1-225.focal-00fdeb8a/

### DIFF
--- a/release.py
+++ b/release.py
@@ -616,7 +616,7 @@ def go(argv):
                         continue
                     # Need to use JENKINS_NODE_TAG parameter for large memory nodes
                     # since it runs qemu emulation
-                    linux_platform_params['JENKINS_NODE_TAG'] = 'large-memory'
+                    linux_platform_params['JENKINS_NODE_TAG'] = 'linux-' + a + '|| large-memory'
 
                 if (NIGHTLY and a == 'i386'):
                     continue


scpeters (approved):

chapulina (approved):
`ign-plugin`'s 1.2.1 release was properly directed to the ARM node using this branch:

https://build.osrfoundation.org/job/ign-plugin-debbuilder/

nuclearsandwich:
Some notes from the run:

An armhf qemu build from March for 1.2.0-1 took 28min on Drogon https://build.osrfoundation.org/job/ign-plugin-debbuilder/614/ compared to a 1.2.1-1 armf native build from today which took 5min 32s https://build.osrfoundation.org/job/ign-plugin-debbuilder/621/


The PR as is will try to find a compatible native host to run on but will fall back to running on a `large-memory` host using qemu. A qemu build from today ran in 29min https://build.osrfoundation.org/job/ign-plugin-debbuilder/622/

Do we want to run arm builds exclusively on native hosts, requiring us to either always run at least one arm host or manually launch one whenever building for ARM or let the current configuration stand and expect to scale out arm hosts when we're doing many coordinated releases?

chapulina:
>  let the current configuration stand and expect to scale out arm hosts when we're doing many coordinated releases?

I think this is the safest for now, considering that not many people know how to scale the hosts and we're often making stable releases on demand.

PR #522, opened 2021-09-28
sdformat.dsl: use default label for brew CI

scpeters:
Use default `osx` label for sdformat brew CI jobs. Currently they are using `osx_mojave`, but there are no more machines in the build farm with that label, since they have had OS upgrades:

* https://build.osrfoundation.org/job/sdformat-ci-sdformat11-homebrew-amd64/

### DIFF
--- a/jenkins-scripts/dsl/sdformat.dsl
+++ b/jenkins-scripts/dsl/sdformat.dsl
@@ -292,10 +292,6 @@ all_branches.each { branch ->
         scm('@daily')
       }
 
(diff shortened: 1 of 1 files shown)


scpeters:
testing:

[Build Status](https://build.osrfoundation.org/job/_dsl_sdformat/1050/) https://build.osrfoundation.org/job/_dsl_sdformat/1050/

azeey (approved):

Issue #554, opened 2021-11-12
Bump in Garden: ign-msgs and dependents


chapulina:
To avoid conflicts, this ticket should be worked on after #545 is closed.

Libraries being bumped:

* `ign-msgs 9`: Various messages will be updated, for example https://github.com/ignitionrobotics/ign-msgs/pull/199, https://github.com/ignitionrobotics/ign-gazebo/issues/1156
    - [x] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437))
    - [x] Update gzdev to use nightlies on CI ([example](https://github.com/ignition-tooling/gzdev/pull/31))
    - [x] homebrew-simulation: create formula and update dependencies ([example](https://github.com/osrf/homebrew-simulation/pull/14230))
    - [x] docs (the collection’s page) ([example](https://github.com/ignitionrobotics/docs/pull/175))

* `ign-fuel-tools 8`: Depends on `ign-msgs`
    - [x] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437))
    - [x] Update gzdev to use nightlies on CI ([example](https://github.com/ignition-tooling/gzdev/pull/31))
    - [x] homebrew-simulation: create formula and update dependencies ([example](https://github.com/osrf/homebrew-simulation/pull/14230))
    - [x] docs (the collection’s page) ([example](https://github.com/ignitionrobotics/docs/pull/175))

* `ign-transport 12`: Depends on `ign-msgs`
    - [x] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
[...]

PR #557, opened 2021-11-17
bump_dependency.bash: improve homebrew support

scpeters:
There were several issues with the auto-generated homebrew formulae in the `bump_dependency.bash` script.

* The major version number in the ign-collection formula was negative (see https://github.com/osrf/homebrew-simulation/pull/1422/commits/509abd322c131e2e31b92a8dc15d101c8fc4eb3f)
* The url replacing logic would replace every url in the file, including the PyYAML url
* It did not remove `revision` numbers
* It was using source commits in the URLs instead of building from the `main` branch, as decided in https://github.com/osrf/homebrew-simulation/pull/1481
* It didn't handle SDFormat properly in a few places

I've used this to open the homebrew-simulation pull requests for https://github.com/ignition-tooling/release-tools/issues/545

### DIFF
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -418,9 +418,11 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
   # remove bottle - TODO: this is only needed for new formulae
   sed -i -e "/bottle do/,/end/d" $FORMULA
   # URL from release to commit - TODO: remove manual step
-  sed -i "s@url.*@url \"$URL\"@g" $FORMULA
+  sed -i "s@^  url.*@  url \"$URL\"@g" $FORMULA
   # SHA - TODO: remove manual step
   sed -i "s/sha256.*/sha256 \"$SHA\"/g" $FORMULA
+  # revision - remove if present
+  sed -i "/^  revision.*/d" $FORMULA
   # version
   sed -i "/ version /d" $FORMULA
   sed -i "/url.*/a \ \ version\ \"${PREV_VER}.999.999~0~`date +"%Y%m%d"`~${SOURCE_COMMIT:0:6}\"" $FORMULA
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -325,7 +325,7 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
   startFromCleanBranch ${BUMP_BRANCH} main
 
   # Check if main branch of that library is the correct version
-  PROJECT_NAME="${LIB//-/_}${VER}"
+  PROJECT_NAME="${LIB_}${VER}"
   PROJECT_NAME="${PROJECT_NAME/ign_/ignition-}"
   PROJECT="project.*(${PROJECT_NAME}"
   if ! grep -q ${PROJECT} "CMakeLists.txt"; then
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -311,9 +311,6 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
   PREV_VER="$((${VER}-1))"
   LIB_UPPER=`echo ${LIB#"ign-"} | tr a-z A-Z`
   ORG=${IGN_ORG}
-  if [ "$LIB" = "sdformat" ]; then
-    ORG=${OSRF_ORG}
-  fi
   BUMP_BRANCH="bump_${COLLECTION}_${LIB}${VER}"
 
   echo -e "${BLUE_BG}Processing [${LIB}]${DEFAULT_BG}"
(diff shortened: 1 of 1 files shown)


release-repo-scripts/bump_dependency.bash
```diff
@@ -57,6 +57,9 @@ VERSION_INPUT=${3}
 ISSUE_NUMBER=${4}
 PREV_COLLECTION=${5}
 DOCS_BRANCH=${6}
+if [[ -z "${DOCS_BRANCH}" ]]; then
+  DOCS_BRANCH=master
+fi
```
adlarkin, line 62: It would probably be worth adding some documentation that `master` is the default branch if `DOCS_BRANCH` is empty. I think this could be mentioned after the `usage` docs: https://github.com/ignition-tooling/release-tools/blob/9f072b3adeda6e844f321e6d7a25d8bd0dfb6a42/release-repo-scripts/bump_dependency.bash#L21-L22

adlarkin (commented):
Changes look okay to me, but I'm not sure how we should test this. Is there a way to test the script locally without opening other PRs? Or, should we just merge this and then see how things go next time the script is used to bump dependencies?

scpeters:
> Is there a way to test the script locally without opening other PRs?

Yes, the script prompts you before committing or opening any PRs. So this can be tested. You can also comment out the `git push` or `gh` command to create pull requests if you want to be extra sure.

release-repo-scripts/bump_dependency.bash
scpeters: https://github.com/ignition-tooling/release-tools/pull/557/commits/605085cbfb4de98f3c6d3ed7d07dac0be615954d

release-repo-scripts/bump_dependency.bash
```diff
@@ -19,10 +19,10 @@
 # Requires the 'gh' CLI to be installed.
 #
 # Usage:
-# $ ./bump_dependency.bash <collection> <library>;<library> <version>;<version> <issue_number> <prev_collection> <docs_branch>
+# $ ./bump_dependency.bash <collection> <library>;<library> <version>;<version> <issue_number> <prev_collection> [<docs_branch>]
 #
 # For example, to bump to `ign-rendering6` and all its dependencies, as well as
-# `sdf12` and its dependencies on fortress:
+# `sdf12` and its dependencies on fortress using the default `master` branch for `docs`:
```
adlarkin, line 25: Does the command below actually use the default `master` branch for `docs`? The command below has `"chapulina/fortress"` as the `docs_branch` argument

scpeters: yeah you're right; I'll fix that

adlarkin:
If I run the script as follows (based on the example command in the script), I get an error:

```
./bump_dependency.bash fortress "ign-rendering;sdformat" "6;12" 428 edifice

...

Processing [ign-rendering6.yaml]
No main or master branch found on ign-rendering6.yaml.
```

It looks like the error comes from this line:
https://github.com/ignition-tooling/release-tools/blob/605085cbfb4de98f3c6d3ed7d07dac0be615954d/release-repo-scripts/bump_dependency.bash#L248

It is true that there is no main/master branch in `ign-rendering6.yaml`: https://github.com/ignition-tooling/gazebodistro/blob/4778f5eb88f9c5f2ac96c86cf2187f3d11e8b806/ign-rendering6.yaml

Am I doing something wrong, or does the script need to be updated?

scpeters:
> If I run the script as follows (based on the example command in the script's header comments), I get an error:
> 
> ```
> ./bump_dependency.bash fortress "ign-rendering;sdformat" "6;12" 428 edifice
> 
> ...
> 
> Processing [ign-rendering6.yaml]
> No main or master branch found on ign-rendering6.yaml.
> ```
> 
> It looks like the error comes from this line:
> 
> https://github.com/ignition-tooling/release-tools/blob/605085cbfb4de98f3c6d3ed7d07dac0be615954d/release-repo-scripts/bump_dependency.bash#L248
> 
> It is true that there is no main/master branch in `ign-rendering6.yaml`: https://github.com/ignition-tooling/gazebodistro/blob/4778f5eb88f9c5f2ac96c86cf2187f3d11e8b806/ign-rendering6.yaml
> 
> Am I doing something wrong, or does the script need to be updated?

I think the script doesn't work once changes. have already been made. I'm not sure there will be an example script we could provide that would always work. You could try figuring out the correct parameters for the bumps in https://github.com/ignition-tooling/release-tools/issues/554

adlarkin:
> You could try figuring out the correct parameters for the bumps in #554

I went ahead and ran tested the script for #554 by running the following command:
```
./bump_dependency.bash garden "ign-msgs" "9" 554 fortress
```

I didn't commit any PRs, but the script successfully made changes to `ign-msgs` and its dependencies. I skimmed through the diffs, and they seemed okay. I think that whoever is going to work on #554 should use this script as a way to test this PR (I will leave a comment on #554 about this).

release-repo-scripts/bump_dependency.bash
scpeters: https://github.com/ignition-tooling/release-tools/pull/557/commits/e6408ec28f13ccacfb178c9814728cc728bb0078

adlarkin (approved):
I think we can merge this and then use it in the next set of version bumps. If any more issues arise when using this script in the next set of bumps, we can always open another PR with more fixes.

PR #559, opened 2021-11-17
Make debbuilds unstable when a missing component is triggering a warning

j-rivero:
This should help to identify when we are missing optional dependencies in the packaging jobs by turning unstable the builds.

Tested here: [Build Status](https://build.osrfoundation.org/job/ign-physics5-debbuilder/523/) with the results of the parser here https://build.osrfoundation.org/job/ign-physics5-debbuilder/523/parsed_console/.

If we merge it now, there would be a good bunch of unstable builds due to different missing packages. The way to solve it would be to inject in the packaging rules files the corresponding `SKIP_$software=True` in the `override_dh_auto_configure` like we are doing for other configurations (i.e: https://github.com/ignition-release/ign-msgs6-release/blob/main/ubuntu/debian/rules#L33-L36).

From the top of my head we would need to patch ign-physics (for DART in Debian) and ign-rendering (for Optix).

### DIFF
--- a/jenkins-scripts/parser_rules/debbuild_missing.parser
+++ b/jenkins-scripts/parser_rules/debbuild_missing.parser
@@ -1,3 +1,3 @@
 warning /^dh_missing: /
 warning /.*binaries-have-file-conflict.*/
-warning /^[[:space:]]*-- Skipping component \[.*\]: Missing dependency.*\./
+warning /^\s*-- Skipping component \[.*\]: Missing dependency.*\./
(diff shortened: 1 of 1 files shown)


scpeters:
this looks like it catches this specific warning. If we prefer narrowly targeting these warnings, then I think this is the way to go. An alternative would be to add the cmake warning parser, but that will probably find lots more than just this one. What do you think?

j-rivero:
> this looks like it catches this specific warning. If we prefer narrowly targeting these warnings, then I think this is the way to go. An alternative would be to add the cmake warning parser, but that will probably find lots more than just this one. What do you think?

Yep, my first intention was to add the CMake warning to it but the debian build environment injects some variables that triggers  some warnings about unused variables I'm not interested into so I just went with this approach that fits well my intentions that are to make visible when we are missing an optional support.

scpeters (approved):

PR #489, opened 2021-07-14
Snapshot 2021.05.12 for ignition vcpkg

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

This PR is related with this https://github.com/ignition-tooling/release-tools/pull/488

https://github.com/ignitionrobotics/ign-gui/pull/250 requires to install `mesa[gles2]`. These two packages are not available in the configured vcpkg tag `2020.11`. We should use a newer version `2021.05.12`.

PR #569, opened 2021-11-22
Fail hard on Windows when a command fails

j-rivero:
Do [not use /b](https://ss64.com/nt/exit.html) on error routines to let Windows kill the whole terminal so the build scripts don't keep running.

Testing without this patch: https://build.osrfoundation.org/job/_test_ign_gui/5/consoleFull#console-section-15 (build is failing on `colcon build` but the script keep going)

Testing with this patch: https://build.osrfoundation.org/job/_test_ign_gui/15/consoleFull#console-section-15 (build is failing as soon as the first problem appear)

### DIFF
--- a/jenkins-scripts/lib/colcon-default-devel-windows.bat
+++ b/jenkins-scripts/lib/colcon-default-devel-windows.bat
@@ -130,4 +130,4 @@ goto :EOF
 
 :error - error routine
 echo Failed with error #%errorlevel%.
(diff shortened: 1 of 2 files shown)


chapulina (approved):

Blast545 (approved):
LGTM

j-rivero:
I'm giving this a few days not to mix the vcpkg change in #489 with this new option.

PR #572, opened 2021-11-24
Try to checkout matching gzdev branch for PR jobs

scpeters:
If a GitHub pull request branch contains `matching_branch/`, then try to checkout that branch from `gzdev`. To simplify checking out other branches, clone gzdev without `--depth 1`.

Part of #564.

Testing:

* https://github.com/ignitionrobotics/ign-fuel-tools/pull/218
* https://github.com/ignition-tooling/gzdev/compare/scpeters/matching_branch/test_msgs9?expand=1
* [Build Status](https://build.osrfoundation.org/job/ignition_fuel-tools-ci-pr_any-ubuntu_auto-amd64/673/) https://build.osrfoundation.org/job/ignition_fuel-tools-ci-pr_any-ubuntu_auto-amd64/673/

Issue #564, opened 2021-11-19
Use matching branch in CI from infrastructure repositories


scpeters:
CI is challenging for PRs that bump version dependencies because they require changes to infrastructure repositories like the following:

* https://github.com/ignition-tooling/gazebodistro
* https://github.com/ignition-tooling/gzdev
* https://github.com/ignition-tooling/release-tools
* https://github.com/osrf/homebrew-simulation

In practice, we often merge the infrastructure PRs first to allow CI to work in the library PR then try to merge the library PR as quickly as possible because other pull requests will be broken. I think we could simplify this by changing our `-ci-pr_any-` builds to check if the library branch name matches a branch name in these infrastructure repositories and use the matching name if so. We could also consider requiring a special string in the branch name (like `branch_match`) to avoid false positives.

scpeters:
A first attempt at this for `gzdev` is in #572

scpeters:
brew CI should be handled by https://github.com/ignition-tooling/release-tools/pull/575

scpeters:
* [X] https://github.com/ignition-tooling/release-tools/pull/572
* [ ] https://github.com/ignition-tooling/release-tools/pull/575
* [ ] Use matching `gzdev` branch in https://github.com/ignition-tooling/action-ignition-ci Ubuntu builds
* [ ] Use matching `homebrew-simulation` branch in https://github.com/ignition-tooling/action-ignition-ci macOS builds
* [ ] Use matching `gazebodistro` branch in Windows builds

PR #575, opened 2021-11-25
Brew CI: try checking out matching branches

scpeters:
Part of #564, similar to #572.

Tested with

* https://github.com/ignitionrobotics/ign-fuel-tools/pull/218
* https://github.com/osrf/homebrew-simulation/compare/scpeters/matching_branch/test_msgs9?expand=1
* [Build Status](https://build.osrfoundation.org/job/ignition_fuel-tools-ci-pr_any-homebrew-amd64/656/) https://build.osrfoundation.org/job/ignition_fuel-tools-ci-pr_any-homebrew-amd64/656/

The `brew pull` command has been deprecated, so that code was replaced with checking for matching branches.

### DIFF
--- a/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
@@ -68,6 +68,7 @@ then
   pushd $(brew --repo osrf/simulation)
   git fetch origin ${ghprbSourceBranch} || true
   git checkout ${ghprbSourceBranch} || true
+  popd
   echo '# END SECTION'
 fi
 
--- a/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
@@ -62,7 +62,7 @@ brew tap osrf/simulation
 echo '# END SECTION'
 
 if [[ -n "${ghprbSourceBranch}" ]] && \
-   [[ "${ghprbSourceBranch}" =~ matching_branch\/ ]]
+   [[ "${ghprbSourceBranch}" =~ ci_matching_branch\/ ]]
 then
   echo "# BEGIN SECTION: trying to checkout branch ${ghprbSourceBranch} from osrf/simulation"
   pushd $(brew --repo osrf/simulation)
--- a/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
@@ -61,9 +61,13 @@ echo '# BEGIN SECTION: setup the osrf/simulation tap'
 brew tap osrf/simulation
 echo '# END SECTION'
 
-if [[ -n "${PULL_REQUEST_URL}" ]]; then
-  echo "# BEGIN SECTION: pulling ${PULL_REQUEST_URL}"
-  brew pull ${PULL_REQUEST_URL}
+if [[ -n "${ghprbSourceBranch}" ]] && \
+   [[ "${ghprbSourceBranch}" =~ matching_branch\/ ]]
+then
+  echo "# BEGIN SECTION: trying to checkout branch ${ghprbSourceBranch} from osrf/simulation"
+  pushd $(brew --repo osrf/simulation)
+  git fetch origin ${ghprbSourceBranch} || true
+  git checkout ${ghprbSourceBranch} || true
   echo '# END SECTION'
 fi
 


jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
```diff
@@ -61,9 +61,14 @@ echo '# BEGIN SECTION: setup the osrf/simulation tap'
 brew tap osrf/simulation
 echo '# END SECTION'
 
-if [[ -n "${PULL_REQUEST_URL}" ]]; then
-  echo "# BEGIN SECTION: pulling ${PULL_REQUEST_URL}"
-  brew pull ${PULL_REQUEST_URL}
```
chapulina, line 66: > The brew pull command has been deprecated

So there's no use for `PULL_REQUEST_URL` anymore? Should it be removed?

scpeters: it's not used in this job, but I think it is used in a downstream job that updates the bottle hash `sha256` values

scpeters: confirmed that it is still in use in the downstream job

* https://github.com/ignition-tooling/release-tools/blob/9bfe8c23e9801a60da815132395626a508cb4860/jenkins-scripts/dsl/brew_release.dsl#L224-L235
* https://github.com/ignition-tooling/release-tools/blob/9bfe8c23e9801a60da815132395626a508cb4860/jenkins-scripts/dsl/brew_release.dsl#L257-L259

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
```diff
@@ -61,9 +61,14 @@ echo '# BEGIN SECTION: setup the osrf/simulation tap'
...
+if [[ -n "${ghprbSourceBranch}" ]] && \
+   [[ "${ghprbSourceBranch}" =~ matching_branch\/ ]]
```
j-rivero, line 65: we need to update the regexp for matching branches accordingly to #572. It would be great if we can avoid the pattern duplication.

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
j-rivero: > confirmed that it is still in use in the downstream job

If I read correctly the DSL code:
```dsl
  parameters { 
        currentBuild() 
          predefinedProp("PULL_REQUEST_URL", "https://github.com/osrf/homebrew-simulation/pull/\${ghprbPullId}") 
}
```

Looks to me like we are overriding `PULL_REQUEST_URL` value to the base URL + `ghprbPullId` instead of using the current one in this script (injected by `currentBuild()`.

If we remove `PULL_REQUEST_URL` from here, is there anyway of launching the job manually? Can that be done without these changes? Maybe is not a supported use case.

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
scpeters: updated in https://github.com/ignition-tooling/release-tools/pull/575/commits/8a9b0c90b0f1f7f42f7a79e6c6c4931ec586dbab

I suppose we could set a variable in `dependencies_archive.sh`, which would work for homebrew and Ubuntu jobs on build.osrfoundation.org, but not on Windows. We could encode the pattern in a file? It seems simpler to just duplicate it for now

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
scpeters: ok, maybe some of this logic is left over from when `release.py` could trigger a bottle build and would pass `PULL_REQUEST_URL`. Now we only start bottle builds from GitHub comments. I'll see if there's something that should be removed

scpeters: > If we remove PULL_REQUEST_URL from here, is there anyway of launching the job manually? Can that be done without these changes? Maybe is not a supported use case.

it's not easy to start the "bottle hash updater job" manually since it needs artifacts from a bottle build

I looked and I don't think there's anything more to be changed

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
j-rivero: For multi-platform support that works on Win and UNIX I can not think in a different option than using a file, yes. Would be great to have this kind of support but by now I'm more worried about having the same regexp in two different files which is a call for problems in the future.

chapulina (approved):
LGTM

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
scpeters: I'll open an issue for this in order to move forward

scpeters: https://github.com/ignition-tooling/release-tools/issues/579

PR #580, opened 2021-12-03
Use ci_matching_branch/ in bump_dependency.bash

scpeters:
This updates the `bump_dependency.bash` script to prefix branch names with `ci_matching_branch/` so that CI should automatically use the matching branch in infrastructure repositories. I'm using this branch with #554, and branch matching currently works for `homebrew-simulation`, so I would expect the homebrew builds to work properly.

The following changes support the matching branch names:

* https://github.com/ignition-tooling/release-tools/commit/861c0229633177b970837a67e0ebf6c68d64a839: use consistent branch names with `ci_matching_branch/` prefix
* https://github.com/ignition-tooling/release-tools/commit/7a0f7369d79d8a2a8efdd87e3a3c7c103f2cc62b: fix sanity check of branch names, which was hard-coded to expect that the names start with `bump_`

The following fixes a bug in fuel-tools branch name replacement:

* https://github.com/ignition-tooling/release-tools/commit/efce3f61afb147e01ec75f09b18e4c5cda717143: reorder the `-` for `_` substitution in the search/replace for collection repositories since it didn't replace `ign-fuel-tools8` with `main` properly

I've also re-ordered the pull requests so that metadata pull requests should be opened first. I think the CI jobs will be more likely to pass with this approach.

### DIFF
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -344,14 +344,14 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
 
     find . -type f ! -name 'Changelog.md' ! -name 'Migration.md' -print0 | xargs -0 sed -i "s ${DEP_LIB}${DEP_PREV_VER} ${DEP_LIB}${DEP_VER} g"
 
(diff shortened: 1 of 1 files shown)


chapulina (approved):
Nice! :+1:

PR #584, opened 2021-12-04
Fix _ handling in bump_dependency for formula

scpeters:
There were some errors in osrf/homebrew-simulation#1704 with `fuel_tools` not properly recognized by the regex. I re-ran with this branch to generate https://github.com/osrf/homebrew-simulation/pull/1704/commits/f31f88251caf8aeeac1c0ffbd99069a9441f3095.

### DIFF
--- a/release-repo-scripts/bump_dependency.bash
+++ b/release-repo-scripts/bump_dependency.bash
@@ -371,7 +371,7 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
 
   # libN
   sed -i -E "s ((${LIB#"ign-"}))${PREV_VER} \1${VER} g" $FORMULA
-  sed -i -E "s ((${LIB_#"ign-"}))${PREV_VER} \1${VER} g" $FORMULA
+  sed -i -E "s ((${LIB_#"ign_"}))${PREV_VER} \1${VER} g" $FORMULA
   # ign-libN -> main
   sed -i "s ${LIB}${PREV_VER} main g" $FORMULA
   # class IgnitionLibN


release-repo-scripts/bump_dependency.bash
```diff
@@ -371,7 +371,7 @@ for ((i = 0; i < "${#LIBRARIES[@]}"; i++)); do
 
   # libN
   sed -i -E "s ((${LIB#"ign-"}))${PREV_VER} \1${VER} g" $FORMULA
-  sed -i -E "s ((${LIB_#"ign-"}))${PREV_VER} \1${VER} g" $FORMULA
+  sed -i -E "s ((${LIB_#"ign_"}))${PREV_VER} \1${VER} g" $FORMULA
```
chapulina, line 374: Cool, it fixes it, but I wonder if there are other problems caused by this. I think `LIB` right now would be something like `ign-gazebo`, and `LIB_` is `ign_fuel_tools`, but the equivalent should be `ign-fuel_tools`.

chapulina (approved):

release-repo-scripts/bump_dependency.bash
scpeters: in the case of fuel tools, `LIB` is `ign-fuel-tools` and `LIB_` is `ign_fuel_tools`. The `#"ign-"` and `#"ign_"` commands strip the prefix so that `fuel-toolsX` and `fuel_toolsX` are the strings that are used for search.

Without this change, the first `sed` command was replacing `fuel-tools7` with `fuel-tools8`, while the second command was trying to replace `ign_fuel_tools7` with `ign_fuel_tools8`, which isn't as useful

Issue #545, opened 2021-11-04
Bump in Garden: ign-rendering and dependents


scpeters:
This is similar to https://github.com/ignition-tooling/release-tools/issues/331

Libraries being bumped:

* ignition-rendering7: needed to support wide-angle cameras, which require new virtual functions in Scene.hh
  * Dependents:
    * ignition-gui7
    * ignition-sensors7
    * ignition-gazebo7
    * ignition-launch6

* ign-rendering7: needed to support wide-angle cameras, which require new virtual functions in Scene.hh
    - [x] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437)) https://github.com/ignition-tooling/release-tools/pull/558
    - [ ] I think this is not needed for ign-rendering, but only for its dependents ~~Update gzdev to use nightlies on CI ([example](https://github.com/ignition-tooling/gzdev/pull/31))~~
    - [x] homebrew-simulation: create formula and update dependencies ([example](https://github.com/osrf/homebrew-simulation/pull/14230)) https://github.com/osrf/homebrew-simulation/pull/1685
    - [x] docs (the collection’s page) ([example](https://github.com/ignitionrobotics/docs/pull/175)) https://github.com/ignitionrobotics/docs/pull/216

* ign-gui7
    - [x] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437))
    - [x] Update gzdev to use nightlies on CI ([example](https://github.com/ignition-tooling/gzdev/pull/31)) https://github.com/ignition-tooling/gzdev/pull/40
[...]

PR #586, opened 2021-12-06
garden view: show main branch CI jobs

scpeters:
I guess we haven't been updating the view as part of the tickets like #554 and #545. This updates the CI job listing to show the relevant `main` branch jobs instead of the Fortress release branches.


chapulina (approved):
One more thing to remember to add to the script

PR #591, opened 2021-12-10
Clean up any possible lock on gitconfig from aborted builds

j-rivero:
Untested but should fix https://github.com/osrf/buildfarmer/issues/257 .

I don't know if there is a better way to avoid the presence of the lock on the aborted builds but this should not cause too much problem.


scpeters (approved):

nuclearsandwich:
In addition to removing the lockfile could we set the actual .gitconfig file to a known state whenever removing it so that we're never left with an invalid gitconfig?

scpeters:
> In addition to removing the lockfile could we set the actual .gitconfig file to a known state whenever removing it so that we're never left with an invalid gitconfig?

the two lines following the `rm` command set the user name and email, so I think we are already doing that

PR #593, opened 2021-12-14
Fix build/ cleanup in colcon Windows build

j-rivero:
Somehow one of the colcon windows build for ign-gazebo failed with problems on the build/ directory https://github.com/ignitionrobotics/ign-gazebo/pull/1246#issuecomment-993714643. 

I think that our check for `KEEP_WORKSPACE` flag is wrong and is not cleaning up the workspaces as expected.

### DIFF
--- a/jenkins-scripts/lib/colcon-default-devel-windows.bat
+++ b/jenkins-scripts/lib/colcon-default-devel-windows.bat
@@ -66,7 +66,7 @@ call %win_lib% :configure_msvc2019_compiler
 echo # END SECTION
 
 echo # BEGIN SECTION: setup workspace
-if defined KEEP_WORKSPACE (
+if not defined KEEP_WORKSPACE (
   IF exist %LOCAL_WS_BUILD% (
     echo # BEGIN SECTION: preclean workspace
     rmdir /s /q %LOCAL_WS_BUILD% || goto :error


Blast545:
Do we have any documentation / standard for that flag to be set? In particular, I recently realized there's a similar flag that to me it seems it's not being used but set everywhere: 
https://github.com/ignition-tooling/release-tools/search?q=IGN_CLEAN_WORKSPACE

https://github.com/ignition-tooling/release-tools/search?q=KEEP_WORKSPACE
Should the usage of those be unified/revised somehow?

j-rivero:
> Do we have any documentation / standard for that flag to be set? In particular, I recently realized there's a similar flag that to me it seems it's not being used but set everywhere: https://github.com/ignition-tooling/release-tools/search?q=IGN_CLEAN_WORKSPACE
> 
> https://github.com/ignition-tooling/release-tools/search?q=KEEP_WORKSPACE Should the usage of those be unified/revised somehow?

That is a good question. They are completely undocumented and unless you edit the jobs manually to include them there is no way to make them work. When I implemented them, they were a debug mechanism that barely is being used so I did not implement the Jenkins parameter that use them to keep the number of options for the user as small as possible.

Blast545 (approved):
LGTM

chapulina (approved):

PR #604, opened 2021-12-23
EOL template: add a task about notifying the community

peci1:
This could be nice as it could actually trigger some people to finish off work they had half-done, instead of letting it time out and having to rebase it for other versions.


chapulina (approved):
Good call :+1: We mentioned it during the community meeting, but a post would have a wider audience.

PR #619, opened 2022-01-05
Remove IGN_PHYSICS_DEPENDENCIES variable

scpeters:
The dependencies are already properly defined in the
packages.apt files in the ign-physics repository.
The IGN_PHYSICS_DEPENDENCIES variable in
dependencies_archive.sh is incorrect, and we'd be
better off without it.

Focal builds are broken: [Build Status](https://build.osrfoundation.org/job/ignition_physics-ci-ign-physics5-focal-amd64/1/) https://build.osrfoundation.org/job/ignition_physics-ci-ign-physics5-focal-amd64/1/

Testing with this branch: [Build Status](https://build.osrfoundation.org/job/ignition_physics-ci-ign-physics5-focal-amd64/2/) https://build.osrfoundation.org/job/ignition_physics-ci-ign-physics5-focal-amd64/2/

### DIFF
--- a/jenkins-scripts/lib/dependencies_archive.sh
+++ b/jenkins-scripts/lib/dependencies_archive.sh
@@ -521,23 +521,6 @@ elif [[ -n "${IGN_GUI_MAJOR_VERSION}" && ${IGN_GUI_MAJOR_VERSION} -eq 3 ]]; then
                         libignition-transport8-dev"
 fi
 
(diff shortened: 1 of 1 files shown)


chapulina (approved):
+1, there's probably much more we can remove too

scpeters:
> +1, there's probably much more we can remove too

yes, there probably is, but I thought I would narrowly scope this change just to fix the broken builds

PR #621, opened 2022-01-05
Add python3-distutils to IGN_MATH_DEPENDENCIES

j-rivero:
Should fix problems on CI in https://github.com/ignitionrobotics/ign-math/pull/360

Tested [Build Status](https://build.osrfoundation.org/job/ignition_math-ci-pr_any-ubuntu_auto-amd64/1190/)

### DIFF
--- a/jenkins-scripts/lib/dependencies_archive.sh
+++ b/jenkins-scripts/lib/dependencies_archive.sh
@@ -374,6 +374,7 @@ fi
 #
 
 IGN_MATH_DEPENDENCIES="libeigen3-dev \\
+                       python3-distutils \\
                        libpython3-dev \\
                        ruby-dev \\
                        swig \\


chapulina:
Thanks! I added that to `ign-math`'s `packages.apt` together with the rest of dependencies in https://github.com/ignitionrobotics/ign-math/pull/360. That's a better place for it since it's used by GitHub actions, tutorials, etc.

scpeters:
> Thanks! I added that to `ign-math`'s `packages.apt` together with the rest of dependencies in [ignitionrobotics/ign-math#360](https://github.com/ignitionrobotics/ign-math/pull/360). That's a better place for it since it's used by GitHub actions, tutorials, etc.

agreed. I proposed moving away from storing dependency information in `dependencies_archive.sh` in #619 since it is easier to express the logic for different versions and OS combinations with the `packages.apt` files

j-rivero:
> agreed. I proposed moving away from storing dependency information in `dependencies_archive.sh` in #619 since it is easier to express the logic for different versions and OS combinations with the `packages.apt` files

ack. I think that we have discussed this in the past, moving dependencies to a more distributed approach could be easier to manage in the short term but it comes with the cost of increasing maintenance effort for changing affecting multiple versions. No silver bullet here.

Do we have a tracking issue to check the status of the migration effort?

chapulina:
> I think that we have discussed this in the past

I think this issue is related, I'll add a comment there: https://github.com/ignition-tooling/release-tools/issues/190

j-rivero:
This is now added in https://github.com/ignitionrobotics/ign-math/blob/ign-math6/.github/ci/packages.apt#L3. Closing.

PR #631, opened 2022-01-27
Define SKIP_usd only for sdformat version 12 and higher

azeey:
This to eliminate warnings in CI for sdformat11 and earlier.


scpeters:
great minds think alike https://github.com/ignition-tooling/release-tools/pull/632

azeey:
Duplicate

scpeters:
you were first though

PR #637, opened 2022-02-09
Use INVALIDATE_DOCKER_CACHE in nightly generation

j-rivero:
There has been problems with recent nightlies that failed due to problems in the upgrade path of packages. Example: [Build Status](https://build.osrfoundation.org/job/ign-gazebo7-debbuilder/186/)

The PR force the `INVALIDATE_DOCKER_CACHE` to run all the nightly builds without cache. Tested in the same node failing before, here: [Build Status](https://build.osrfoundation.org/job/ign-gazebo7-debbuilder/189/)


Blast545:
@j-rivero Bad news on this fron, it seems exporting this variable wasn't enough to address the issue. We have these today:

https://build.osrfoundation.org/job/ign-launch6-debbuilder/195
https://build.osrfoundation.org/job/ign-gazebo7-debbuilder/193

And according to the log, the same problem still happens, as the dependencies being fetched from a previous build.

PR #641, opened 2022-02-11
Use xz to compress .deb packages on jammy

j-rivero:
The problem with the default compression to zstd on Jammy for .deb packages and our old version of reprepro can be workarounded by keeping compression in xz.

dpkg-deb does not expose the compression option for the internal artifacts inside .deb (the options available seems to affect only to the tarballs not in the .deb). The workaround is a bit tricky since we need to create a wrapper that forces this option, see https://alioth-lists-archive.debian.net/pipermail/devscripts-devel/2014-May/002232.html

Tested here: https://build.osrfoundation.org/job/ign-cmake2-debbuilder/1343/
Upoaded just fine here: https://build.osrfoundation.org/job/repository_uploader_packages/22181/


j-rivero:
Need to merge it to produce packages

PR #643, opened 2022-02-13
Retry the packages installation in nightlies if failed

j-rivero:
The PR implements the mechanism to repeat the execution of mk-build-deps to help with problems in nightlies. This should cover possible race conditions between different packages being built and uploaded at the same time.

Together with #642, this should help to diagnose the problems on the nightly generation.

Tested: [Build Status](https://build.osrfoundation.org/job/ign-gazebo7-debbuilder/202/)


scpeters (approved):
let's try it

PR #647, opened 2022-02-15
Native docker platform support for armhf on Jammy

j-rivero:
There is no `osrf/jammy_armhf` image so @nuclearsandwich suggested to give it a try to the native platform docker support. That worked just fine: [Build Status](https://build.osrfoundation.org/job/ign-math6-debbuilder/1365/)

@nuclearsandwich commented offline:

> One thing I noticed when doing manual testing, the vanilla docker-build command was not platform aware. So using FROM --platform=linux/armhf ubuntu:jammy would work the first time but if I tried to build or run a different platform image for ubuntu:jammy it would keep using the armhf one until I ran docker-rmi ubuntu:jammy It may be worth testing whether that same host will correctly build an arm64 or amd64 package as well. And you may need to switch to the buildx builder to get it working reliably unless you want to untag or remove your base image before each build. https://docs.docker.com/buildx/working-with-buildx/

I still need to address the possible problems.

### DIFF
--- a/jenkins-scripts/docker/lib/docker_generate_dockerfile.bash
+++ b/jenkins-scripts/docker/lib/docker_generate_dockerfile.bash
@@ -97,8 +97,17 @@ case ${ARCH} in
        FROM_VALUE=${ARCH}/${LINUX_DISTRO}:${DISTRO}
      fi
      ;;
-   'armhf' | 'arm64' )
-       FROM_VALUE=osrf/${LINUX_DISTRO}_${ARCH}:${DISTRO}
+   'armhf')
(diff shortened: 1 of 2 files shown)


nuclearsandwich (commented):
This looks reasonable so far!

j-rivero:
Tested here, [Build Status](https://build.osrfoundation.org/job/ign-math6-debbuilder/).

j-rivero:
> Tested here, [Build Status](https://build.osrfoundation.org/job/ign-math6-debbuilder/).

The buildx changes in this PR went just fine, the autopkgtest test is failing probably because we are rebuilding the same packages already in the repository.

PR #649, opened 2022-02-17
Use docker-jammy TAG in jenkins for Jammy jobs

j-rivero:
The jammy docker image requires updated docker versions to handle it properly. The xenial nodes in the buildfarm are affected and can not be updated easily. The focal nodes, after being updated to version greater than 20.10.07 to deal with different problems on jammy including the clone3 syscall.

This change should be temporary until we update all the nodes.


j-rivero:
Need to fix jammy nightlies. Merging.

PR #650, opened 2022-02-24
Use a special node tag for processing nightlies

j-rivero:
Part of https://github.com/ignition-tooling/release-tools/issues/644 as suggested by @scpeters.

We have Bionic, Focal and Jammy, I've setup temporary the tags in docker (need permanent configuration) as follow:
 * optimus: focal
 * r2d2: bionic
 * linux-ip-172-30-1-216: jammy (not large-memory but can probably handle the existing jammys)

### DIFF
--- a/release.py
+++ b/release.py
@@ -623,6 +623,15 @@ def go(argv):
                 if (NIGHTLY and a == 'i386'):
                     continue
 
+                # control nightly generation using a single machine to process
+                # all distribution builds to avoid race conditions. Note: this
+                # assumes that large-memory nodes are beind used for nightly
+                # tags.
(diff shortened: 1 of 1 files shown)


scpeters (approved):

j-rivero:
> linux-ip-172-30-1-216: jammy (not large-memory but can probably handle the existing jammys)

At some point the node got disconnected and the label was lost. It needs infra deployment, will do it today.

azeey:
I ran into an issue today trying to test a nightly build using `release.py` for Ionic. The `JENKINS_NODE_TAG`  was being set to `linux-nightly-jammy`, but we currently only have the label `linux-nightly-focal` on the node designated to run nightlies.

PR #659, opened 2022-03-10
gazebo.dsl: disable windows builds

scpeters:
The builds are failing due to an old version of tbb
in the zipped dependency archive. See osrf/gazebo#3174
for more information.


chapulina (approved):

scpeters:
now that I think about it; I only wanted to disable gazebo11 builds at the moment, since only gazebo11 has the conda CI and problem with tbb. I'm not ready to merge this yet

scpeters:
I think I've found a fix for the windows builds in https://github.com/osrf/gazebo/pull/3157, so this won't be needed

PR #662, opened 2022-03-11
Update vcpkg-snapshot to 2022.02.23

j-rivero:
Follow up and superseed https://github.com/ignition-tooling/release-tools/pull/657. Originally coming from https://github.com/ignitionrobotics/ign-common/pull/292#issuecomment-1039738233 , problems with gdal in some of the Windows nodes.

We can not install gdal again from previous versions of vcpkg because some of the files have been removed from the mirrors since new versions are out. This PR updates vcpkg to the latest snapshot available. This is going to bump many versions at once, including ogre to 1.12.9, ffmpeg to 4.4.1, etc.

The testing job for compiling all ignition family up to main on[ ign-gazebo seems fine](https://build.osrfoundation.org/job/_vcpkg_testing_snapshot/113/). Probably all the update work done in Brew and all the patched for Windows sent by conda-forge contributors make it to work out-of-box.

There is a problem with a hash in zeromq that I patched in e0505323d3d0d311e0e63d7b91d7d14b4e958221.

I also needed to update our check on the vcpkg tag being used since it failed when multiple tags were in the same commit sha. 48084de907390aea3c2679922213509a5ff900a8.

//cc @Blast545

### DIFF
--- a/jenkins-scripts/lib/windows_env_vars.bat
+++ b/jenkins-scripts/lib/windows_env_vars.bat
@@ -15,6 +15,6 @@ set VCPKG_CMD=%VCPKG_DIR%\vcpkg.exe
 set VCPKG_CMAKE_TOOLCHAIN_FILE=%VCPKG_DIR%/scripts/buildsystems/vcpkg.cmake
 if NOT DEFINED VCPKG_SNAPSHOT (
   :: see https://github.com/microsoft/vcpkg/releases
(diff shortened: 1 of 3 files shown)


Blast545:
`vcpkg` :sob:

Blast545:
`vcpkg` :sob:

Blast545 (approved):
Makes sense to me :rocket:

PR #663, opened 2022-03-11
Remove docker-jammy jenkins label, not needed anymore

j-rivero:
All jenkins nodes were updated to Focal and docker versions should be modern enough to handle jammy. Removing the workaround of `docker-jammy` from release.py.


j-rivero:
This is causing problems on releasing, merging.

PR #680, opened 2022-03-30
Allow CI Workflow jobs to run concurrently.

nuclearsandwich:
At the moment there is a backlog of ign_gazebo-pr-win jobs due in part to the fact that only one may run at a time.

I am not sure if there are some job types which should not be allowed to run conditionally but decided to start by just opening the floodgates in case it was just that easy.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/GenericCompilation.groovy
+++ b/jenkins-scripts/dsl/_configs_/GenericCompilation.groovy
@@ -41,6 +41,7 @@ class GenericCompilation
 
      job.with
      {
+        concurrentBuild(true)
         properties {
           priority 300
         }
--- a/jenkins-scripts/dsl/_configs_/OSRFCIWorkFlowMultiAnyGitHub.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFCIWorkFlowMultiAnyGitHub.groovy
@@ -31,6 +31,7 @@ class OSRFCIWorkFlowMultiAnyGitHub
   {
     job.with
     {
+      concurrentBuild(true)
       logRotator {
         numToKeep(25)
       }
--- a/jenkins-scripts/dsl/_configs_/OSRFCIWorkFlowMultiAnyGitHub.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFCIWorkFlowMultiAnyGitHub.groovy
@@ -31,7 +31,6 @@ class OSRFCIWorkFlowMultiAnyGitHub
   {
     job.with
     {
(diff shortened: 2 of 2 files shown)


azeey (approved):
> I am not sure if there are some job types which should not be allowed to run conditionally 

The only thing I can think of for not allowing jobs to run concurrently is to prevent consecutive instances of a long-running job from hogging all available nodes. But I'd say we can disable this if that really happens.

j-rivero (changes requested):
I think that the intention is correct about declaring concurrent builds.  The `OSRFCIWorkFlowMultiAnyGitHub.groovy` class is used to generate workflows to trigger the different platforms at the same time. I think that it is largely unused nowadays and GitHub is directly linked to `*-ci-pr_any-*` jobs directly (what has different advantages).

If we intent to expand the goal to all kind of compilation (integrated with GitHub or no) in all platforms, we can consider to the `GenericCompilation.groovy` class instead.

nuclearsandwich:
> The `OSRFCIWorkFlowMultiAnyGitHub.groovy` class is used to generate workflows to trigger the different platforms at the same time.

Ah, I misread the ignition DSL. It looked like that was a rollup config for creating each job rather than the job which launches the rest.

I'm not sure which layer is best to enable this on. It does seem like we should be able to scale out any CI/PR jobs by default since they ought not be conflicting with the possible exception of per-host resources like GPUs but that should be handled per node since those conflicts can arise across different jobs as well as builds of the same job.

Is there any documentation that describes the inheritance hierarchy for the different DSL classes? I trust your advice that GenericCompilation.groovy is the place to add this but documentation would have helped me find that answer on my own.

j-rivero:
> I'm not sure which layer is best to enable this on. It does seem like we should be able to scale out any CI/PR jobs by default since they ought not be conflicting with the possible exception of per-host resources like GPUs but that should be handled per node since those conflicts can arise across different jobs as well as builds of the same job.

build.o.o works under the assumption that CI agents only provide one executor. We can consider to increase this at some point but now the conflicts with per-host resources don't apply (if I'm not missing something).

> Is there any documentation that describes the inheritance hierarchy for the different DSL classes? I trust your advice that GenericCompilation.groovy is the place to add this but documentation would have helped me find that answer on my own.

I'm afraid that there is no documentation on DSL classes. My reasoning was: what kind of jobs do we want to parallelize in different nodes? Mostly everything in different platforms. `GenericCompilation` is one of the most cross-platforms ones and particularly affecting building ones.

j-rivero (approved):

PR #683, opened 2022-04-05
Release table: add jammy and improve nightly

chapulina:
* Add Jammy for Fortress and Garden
* Remove Bionic for Garden
* For nightlies, don't look up architectures that don't have them

image

image

### DIFF
--- a/terminal-dashboard/table.bash
+++ b/terminal-dashboard/table.bash
@@ -43,9 +43,12 @@ for LIB in $(get_libraries_by_collection "${COLLECTION}" ); do
   for DISTRO in "${DISTROS[@]}"
   do
     if [[ $DISTRO == "ubuntu" ]]; then
-      VERS=( "bionic" "focal" )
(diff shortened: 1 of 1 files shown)


terminal-dashboard/table.bash
```diff
@@ -59,6 +69,8 @@ for LIB in $(get_libraries_by_collection "${COLLECTION}" ); do
 
         if [[ $ARCH == "i386" && $VER == "focal" ]]; then
```
j-rivero, line 70: There won't be more i386 Ubuntu support by now so we can turn this into: 
```suggestion
        if [[ $ARCH == "i386" && $VER != "bionic" ]]; then
```
and remove the new `elif` for jammy.

j-rivero (approved):
Minor suggestion. Looks good.

terminal-dashboard/table.bash
chapulina: This applies for Debian too, so we need to account for `buster` somewhere

PR #687, opened 2022-04-06
Bumps in garden : ign-utils2

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/685


chapulina (approved):

chapulina:
This is safe to merge already. The `ign-utils1` branch is not having any active changes that we'd need in a new nightly.

PR #690, opened 2022-04-07
Use BUILD_DOCS=OFF for ign-gazebo on Jammy

j-rivero:
doxygen is buggy for ign-gazebo on Jammy https://github.com/ignitionrobotics/ign-gazebo/issues/1409

Twin of CI change done per https://github.com/ignitionrobotics/ign-gazebo/pull/1418#issuecomment-1090710034


chapulina:
I think this may not be necessary, I see this build passed using `master`:

https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/8426/consoleFull

j-rivero:
> I think this may not be necessary, I see this build passed using `master`:

If it is not needed, let's close it. Reopen if needed.

PR #691, opened 2022-04-08
ign-tools doesn't depend on ign-cmake

chapulina:
This was probably added back when we attempted to add `ign-cmake` as a dependency, and it wasn't reverted when we failed.

I noticed it because it's making `ign-cmake2` and `ign-cmake3` be installed side-by-side as we bump libraries in

* https://github.com/ignition-tooling/release-tools/issues/685


methylDragon (approved):
Yep, ign-tools doesn't have a ign-cmake reference in its CMakeLists.txt. I don't expect this to break anything else since it's an extraneous dependency at this point...

scpeters (approved):

j-rivero (approved):

PR #688, opened 2022-04-07
Fix GZDEV_PROJECT_NAME for ign-utils2

chapulina:
Needed by 
* https://github.com/ignition-tooling/release-tools/issues/685
* https://github.com/ignitionrobotics/ign-utils/pull/49

Oops, that was probably bad copy-paste. See how other libs are correct:

https://github.com/ignition-tooling/release-tools/blob/54d09aeadb420e6f9bbb7836fa929cda5dfd1788/jenkins-scripts/docker/ign_math-compilation.bash#L36

https://github.com/ignition-tooling/release-tools/blob/54d09aeadb420e6f9bbb7836fa929cda5dfd1788/jenkins-scripts/docker/ign_msgs-compilation.bash#L35

Without this, `gzdev` is not enabling nightlies for `ign-utils2` ubuntu builds

PR #692, opened 2022-04-08
Fix GZDEV_PROJECT_NAME for ign-plugin2

chapulina:
* Same as #688 
* Needed by https://github.com/ignition-tooling/release-tools/issues/685


chapulina:
Merging to unblock Garden bumps, it's trivial

Issue #685, opened 2022-04-06
Bumps in Garden: ign-cmake3, ign-utils2, ign-tools2, ign-plugin2


methylDragon:
These bumps are being done to prepare for the migration from `ign` to `gz`.

## Libraries being bumped
* `ign-cmake3`:
    - [ ] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437))
    - [x] homebrew-simulation: create formula and update dependencies ([example](https://github.com/osrf/homebrew-simulation/pull/14230))
    - [x] docs (the collection’s page) ([example](https://github.com/ignitionrobotics/docs/pull/175))

* `ign-utils2`:
    - [x] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437))
    - [x] homebrew-simulation: create formula and update dependencies ([example](https://github.com/osrf/homebrew-simulation/pull/14230))
    - [x] docs (the collection’s page) ([example](https://github.com/ignitionrobotics/docs/pull/175))

* `ign-tools2`:
    - [ ] Source code of all downstream libraries ([example](https://github.com/ignitionrobotics/ign-transport/pull/149))
    - [x] Release repositories of all downstream libraries ([example](https://github.com/ignition-release/ign-fuel-tools7-release/pull/1))
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/ignition-tooling/gazebodistro/pull/42))
    - [x] Build nightlies from the `main` branch ([example](https://github.com/ignition-tooling/release-tools/pull/437))
[...]

PR #694, opened 2022-04-08
Bumps in garden : ign-plugin2

methylDragon:
See #685


chapulina (approved):

chapulina:
Safe to merge, `ign-plugin1` doesn't have active development so we don't need its nightlies.

PR #701, opened 2022-04-13
Perform deprecation tocks when a main branch is created

chapulina:
We often forget to execute them. This item should help


j-rivero (approved):
good addition

PR #704, opened 2022-04-15
Deprecate Dependencies Archive for ign-tools installs

methylDragon:
See: https://github.com/ignition-tooling/release-tools/issues/685

This fixes the issue where ignition-tools gets installed alongside ignition-tools2 for versions that use it.

Also, this is a partial deprecation of the dependencies_archive in general.


chapulina (approved):
LGTM! It would be good to run the DSL once and run a test build just to make sure that the syntax is all good

methylDragon:
Green DSL generate: https://build.osrfoundation.org/job/_dsl_ignition_collection/500/
Successful build using the RTools branch: https://build.osrfoundation.org/job/ignition_msgs-ci-pr_any-ubuntu_auto-amd64/847/console

PR #702, opened 2022-04-14
Use colcon auto major version for ign-tools

chapulina:
* Needed by https://github.com/ignitionrobotics/ign-tools/pull/82
* Part of https://github.com/ignition-tooling/release-tools/issues/685

I checked that `jenkins-scripts/tools/detect_cmake_major_version.py` correctly returns `1` for the `ign-tools1` branch and `2` for the branch on the PR above.

Build with this branch: [Build Status](https://build.osrfoundation.org/job/ign_tools-pr-win/44/)

PR #706, opened 2022-04-15
ign-tools auto version

chapulina:
I broke CI for `ign-tools1` on #702 because now the package ends up being `ignition-tools1` when it should be `ignition-tools`.

I moved the logic around to support this specific use case. The `COLCON_AUTO_MAJOR_VERSION` variable's meaning is a bit changed now, but grepping through the code it looks like `ign-tools` is the only one setting it to false anyway.

I don't know if the syntax is correct 🙃 

Test builds:

* `ign-tools1`: [Build Status](https://build.osrfoundation.org/job/ign_tools-pr-win/47/)
* `ign-tools2`: [Build Status](https://build.osrfoundation.org/job/ign_tools-pr-win/48/)

### DIFF
--- a/jenkins-scripts/lib/colcon-default-devel-windows.bat
+++ b/jenkins-scripts/lib/colcon-default-devel-windows.bat
@@ -29,8 +29,8 @@ set LOCAL_WS_BUILD=%WORKSPACE%\build
 @if "%COLCON_AUTO_MAJOR_VERSION%" == "" set COLCON_AUTO_MAJOR_VERSION=false
 
 setlocal ENABLEDELAYEDEXPANSION
-if "%COLCON_AUTO_MAJOR_VERSION%" == "true" (
-   for /f %%i in ('python "%SCRIPT_DIR%\tools\detect_cmake_major_version.py" "%WORKSPACE%\%VCS_DIRECTORY%\CMakeLists.txt"') do set PKG_MAJOR_VERSION=%%i
+for /f %%i in ('python "%SCRIPT_DIR%\tools\detect_cmake_major_version.py" "%WORKSPACE%\%VCS_DIRECTORY%\CMakeLists.txt"') do set PKG_MAJOR_VERSION=%%i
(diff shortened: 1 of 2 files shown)


jenkins-scripts/lib/colcon-default-devel-windows.bat
```diff
@@ -29,8 +29,8 @@ set LOCAL_WS_BUILD=%WORKSPACE%\build
 @if "%COLCON_AUTO_MAJOR_VERSION%" == "" set COLCON_AUTO_MAJOR_VERSION=false
 
 setlocal ENABLEDELAYEDEXPANSION
-if "%COLCON_AUTO_MAJOR_VERSION%" == "true" (
-   for /f %%i in ('python "%SCRIPT_DIR%\tools\detect_cmake_major_version.py" "%WORKSPACE%\%VCS_DIRECTORY%\CMakeLists.txt"') do set PKG_MAJOR_VERSION=%%i
+for /f %%i in ('python "%SCRIPT_DIR%\tools\detect_cmake_major_version.py" "%WORKSPACE%\%VCS_DIRECTORY%\CMakeLists.txt"') do set PKG_MAJOR_VERSION=%%i
+if "%COLCON_AUTO_MAJOR_VERSION%" == "true" || "%PKG_MAJOR_VERSION%" != "1" (
    set COLCON_PACKAGE=%COLCON_PACKAGE%!PKG_MAJOR_VERSION!
```
scpeters, line 34: maybe the problem is this line? perhaps we should be trying to detect the cmake project name rather than the major version?

chapulina: I was trying to keep this generic, but in the end of the day this logic seems like it's here only for `ign-tools`, so we could hardcode the name here. Is this what you mean?

scpeters:
the underlying issue has already been fixed by https://github.com/ignition-tooling/release-tools/pull/713

PR #708, opened 2022-04-15
Depreciate Dependencies Archive for ign-rendering4 installs

methylDragon:
See: https://github.com/ignition-tooling/release-tools/issues/685


chapulina (commented):
👍🏽

Issue #698, opened 2022-04-11
Rename Ignition to Gazebo


chapulina:
See the [announcement](https://community.gazebosim.org/t/a-new-era-for-gazebo/1356). This issue is meant to capture the technical tasks involved in the migration.

- [ ] Transfer GitHub orgs and repos
    - [ ] ignitionrobotics ➡️ gazebosim
    - [ ] ignition-release ➡️ gazebo-release
    - [ ] ignition-forks ➡️ gazebo-forks
    - [ ] ignition-tooling ➡️ gazebo-tooling
    - [ ] Source repos
         * Archived repos keep their names if that's easier
         * s/ign/gz
         * s/gazebo/sim
    - [ ] Release repos
        * archived repos keep their names if that's easier
        * release repos for Citadel and Fortress should probably keep their names if that's easier
        * all other release repos: s/ign/gz
    - [ ] Tooling repos
        * s/ign/gz
        * s/ignition/gz
        * Most won't change

- [ ] C++ namespaces from `ignition` to `gz`
- [ ] Directories from `ignition` to `gz`
- [ ] Header guards
- [ ] Environment variables
- [ ] SDF `ignition:` namespaces
- [ ] Icons on the UI
- [ ] CMake project names (`ignition` to `gz`)
- [ ] Debian packages (`ignition` to `gz`)
- [ ] rosdep keys

chapulina:
> If the ign tool in https://github.com/ignitionrobotics/ign-tools is renamed in gz, there is the possibility of a collision with the Gazebo Classic gz tool. Do you have any idea on how we can handle that?

This is something we've been discussing. Some possible outcomes:

* Gazebo Garden+ and Gazebo Classic can't be installed side-by-side. Garden and Gazebo 11's support timelines only overlap for 2 years and their only officially supported platform in common is Ubuntu Focal. This may hurt people's migration efforts from Classic to Garden, but they have other 2 LTS versions to target: Citadel and Fortress.

* Move the `gz` binary from Gazebo classic to a separate debian package, so `gazebo11` and `libgazebo11-dev` wouldn't collide with Garden, only the new "`gazebo11-gz`" package would. Only a small fraction of users use that tool, so most users shouldn't be affected. But this may be a breaking change to a stable version, which we should avoid. This option isn't available to the packages that are already upstream on Ubuntu.

[...]

PR #709, opened 2022-04-15
Completely rely on packages.apt from Garden

chapulina:
Building on top of @methylDragon 's work in:

* https://github.com/ignition-tooling/release-tools/pull/704
* https://github.com/ignition-tooling/release-tools/pull/705
* https://github.com/ignition-tooling/release-tools/pull/707
* https://github.com/ignition-tooling/release-tools/pull/708

This PR completely skips all dependency resolution for Garden versions. We should probably be able to change this all the way back to Citadel, but for now sticking to Garden to be quicker.

Builds have been installing more than they needed for a while, but it hasn't been a problem until now because all libraries could be installed side-by-side. But while we work on #698, `ign-tools2` will be conflicting with `ign-tools1` for a while. So it's especially important that we only install what we need, as described on `packages.apt`.

➡️ I recommend hiding whitespace changes for easier review

### DIFF
--- a/jenkins-scripts/lib/dependencies_archive.sh
+++ b/jenkins-scripts/lib/dependencies_archive.sh
@@ -384,18 +384,21 @@ fi
 # IGNITION
 #
 
-if [[ ${IGN_MATH_MAJOR_VERSION} -eq 4 ]]; then
-  IGN_MATH_DEPENDENCIES="libeigen3-dev \\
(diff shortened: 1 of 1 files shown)


chapulina:
YOLO merging to unblock https://github.com/ignition-tooling/release-tools/issues/685

🙏🏽 🤞🏽

PR #677, opened 2022-03-25
Script to generate release summaries

chapulina:
Generate markdown summaries to be posted on Community posts like this: [New Ignition releases 2022-03-01 (Citadel, Edifice, Fortress)](https://community.gazebosim.org/t/new-ignition-releases-2022-03-01-citadel-edifice-fortress/1313).

Example:

```
$ bash release_summary.bash transport 10.1.0 10.2.0
---------------------------------
NAME_FOR_TAGS: ignition-transport
NAME_FOR_REPO: ign-transport
NAME_FOR_BRANCH: ign-transport
NAME_FOR_TITLE: Ignition Transport
Previous version: 10.1.0
New version: 10.2.0
Major version: 10
---------------------------------

Fetching origin
remote: Enumerating objects: 61, done.
remote: Counting objects: 100% (59/59), done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 22 (delta 16), reused 20 (delta 15), pack-reused 0
Unpacking objects: 100% (22/22), 2.93 KiB | 38.00 KiB/s, done.
From ssh://github.com/ignitionrobotics/ign-transport
 * [new branch]        chapulina/11_to_12 -> origin/chapulina/11_to_12
   b4473a56..b4709257  ign-transport11    -> origin/ign-transport11
Branch 'ign-transport10' set up to track remote branch 'ign-transport10' from 'origin'.
Switched to a new branch 'ign-transport10'
From ssh://github.com/ignitionrobotics/ign-transport
 * branch              ign-transport10 -> FETCH_HEAD
Already up to date.




## Ignition Transport 10.2.0

## Changelog

[Full changelog](https://github.com/ignitionrobotics/ign-transport/blob/ign-transport10/Changelog.md)


1. Use exec instead of popen to run ign-launch binary
    * [Pull request #300](https://github.com/ignitionrobotics/ign-transport/pull/300)

1. Focal CI: static checkers, doxygen linters, compiler warnings
    * [Pull request #298](https://github.com/ignitionrobotics/ign-transport/pull/298)

1. Add option to output messages in JSON format
    * [Pull request #288](https://github.com/ignitionrobotics/ign-transport/pull/288)

1. Remove no username error messages
    * [Pull request #286](https://github.com/ignitionrobotics/ign-transport/pull/286)

1. Documented the default value of `IGN_PARTITION`
    * [Pull request #281](https://github.com/ignitionrobotics/ign-transport/pull/281)

[...]

PR #712, opened 2022-04-25
Script for Header Migration

methylDragon:
# Description
This is related to https://github.com/ignition-tooling/release-tools/issues/711
And is an extended scripted version of https://github.com/ignitionrobotics/ign-utils/pull/51

It should:
- [x] Move header files to `gz/*` and add redirect headers to `ignition/*`
- [x] Replace all `#include <ignition/*>` statements in header files with `#include <gz/*>`
- [x] Create the associated `Export.hh` and `<lib>.hh` header files
- [x] Replace all `#include <ignition/*>` statements in src, test, and examples source code with `#include <gz/*>`

# Examples
With the script placed into the root of [`ign-utils`](https://github.com/ignitionrobotics/ign-utils), on the main branch and an empty `header_migration` branch, run `./header_migration_script.sh`

PR generated from this script: https://github.com/ignitionrobotics/ign-utils/pull/53

# Further Work
### Generated Files
This script can't handle generated files because it has no knowledge of them, and to get that functionality, it'll need to be able to intelligently go up and down the directory tree, looking into CMake files and the like.

There is too much variance between the packages to do this quickly (and robustly), so for those cases, users will have to manually do the migration. These are some of the common edge cases:
- `Export.hh` and `<lib>.<lib.hh` files have to be manually added since they're usually dynamically generated by `ign-cmake`
  - [Export Example](https://github.com/ignitionrobotics/ign-math/pull/413/commits/edc04863d2608ee72bff07b68f72e52c994c384e) | [lib Example](https://github.com/ignitionrobotics/ign-math/pull/413/commits/edc04863d2608ee72bff07b68f72e52c994c384e#diff-85190fe279f22a6d7e149344c203113fbcc0dedf2a2442084497cdb8ab6952f7)
- And also manually installed in `CMakeLists.txt` files (that usually have to be added)
  - Sometimes, whole migrated packages have to have extra CMake files added to ensure that they install properly
  - [Example](https://github.com/ignitionrobotics/ign-math/pull/413/commits/b6da54841c213f6a137acd3adc1e966e73eafd8e)

Typically, the way you'd surface these issues is, **once you've done the migration with the script**, install the unmigrated version alongside the migrated version (across two workspaces), then use [meld](https://meldmerge.org/) to run a directory diff on:
- The top level directory: To check if the appropriate **redirection headers** are installed!
  - `migrated_ws/install/ignition-<lib>/include/ignition/<lib>` vs;
  - `unmigrated_ws/install/ignition-<lib>/include/ignition/<lib>`
- The target directory: Tto check if the **migrated** headers are installed!
  - `migrated_ws/install/ignition-<lib>/include/ignition/<lib>/gz` vs;
  - `unmigrated_ws/install/ignition-<lib>/include/ignition/<lib>/ignition`

### Greedy Matching
There are a couple of instances where the script will match greedily and change something that shouldn't be changed. This is because I am erring on the side of editing something (so it is visible to a user) rather than missing something.

I think the script gets it right in the ballpark of 90% correct, but there are about 10% of cases that are false positive matches that require users to manually discard changes for.

To aid in this, the script has multiple stopping and review points for a user to open the git diffs on a program like GitKraken to double-check any changes made.

### Already Migrated Files
(This is sort of a variant of greedy matching)

Be aware that the script will greedily convert already migrated `gz` headers into redirection headers! Please check to make sure it doesn't do that!

## Unrelated To This PR
- [ ] Remove the REPLACE_IGNITION_INCLUDE_PATH parameter override once its no longer needed
- [ ] Add using namespace ignition to the ignition/ headers
- [ ] Add portable #warning equivalent to ignition/ headers
- [ ] Replace all #include <ignition/*> statements in downstream code
- [ ] Change ign-cmake3's default value of PROJECT_INCLUDE_DIR from ignition/${IGN_DESIGNATION} to gz/${IGN_DESIGNATION} (or use ${GZ_DESIGNATION} if that has been defined) in [IgnConfigureProject.cmake](https://github.com/ignitionrobotics/ign-cmake/blob/main/cmake/IgnConfigureProject.cmake#L91)


scpeters:
are you planning to match the pattern of commits used in https://github.com/ignitionrobotics/ign-utils/pull/51? I think it's helpful to be strategic about the commits because the overall diff is very hard to read

methylDragon:
The script is currently in a state where it reproduces all the changes made by https://github.com/ignitionrobotics/ign-utils/pull/51 except making the `Export.hh` and `utils.hh`!

I'll be implementing that (though the `utils.hh` will require command line input to resolve properly.)

Then I'll also work on replacing the relevant statements in `src`, `test`, and `examples`. (This will mean I'll also have to migrate the headers in `cli/include`, which I deliberately skipped for now so I could get the script doing what was in https://github.com/ignitionrobotics/ign-utils/pull/51 first.

scpeters:
I just noticed the following header file with `Ignition` in the file name. We'll probably have to deprecate that manually

* https://github.com/ignitionrobotics/ign-utils/blob/ign-utils1/cli/include/ignition/utils/cli/IgnitionFormatter.hpp

methylDragon:
> I just noticed the following header file with `Ignition` in the file name. We'll probably have to deprecate that manually
> 
> * https://github.com/ignitionrobotics/ign-utils/blob/ign-utils1/cli/include/ignition/utils/cli/IgnitionFormatter.hpp

I'll write functionality to migrate the file, but not the class name. Let's do class and variable name migration in a separate set of PRs so things are cleaner?

methylDragon:
> there's a lot happening so it's a bit hard to follow this script. I would imagine the first commit could be done with code like the following:
> 
> ```
> git checkout -b migrate_ign_headers
> 
> # First commit: rename folders
> git mv include/ignition include/gz
> for d in $(ls -d */include/ignition)
> do
>   git mv ${d} ${d}/../gz
> done
> git commit -a --signoff -m "git mv include/ignition include/gz"
> ```

I was using `git mv` originally, just that I ended up running `git reset` to get the diffs to show up (since they'll only show up on unstaged changes.)

That is no longer the case, and history should be preserved!

https://github.com/ignition-tooling/release-tools/pull/712/commits/daad715a0277002fc9426ec3e1fec0b256ac0fcb
https://github.com/ignition-tooling/release-tools/pull/712/commits/0a46de18d5d7b44d317267f05fa83ef9e6c4d726

j-rivero (commented):
Seems like the script is in active development. If that is the case please mark it as Draft.

Before looking at the code, the script is placed under `jenkins-scripts/tools` but I believe that it does not have anything to do with Jenkins. We can create a new subdir in the root of the repo named something like `code-repo-scripts` to host the scripts designed to work with repository code. My intention is to move `source_changelog` into it but that will require a different PR.

jenkins-scripts/tools/header_migration_script.sh
```diff
@@ -0,0 +1,500 @@
+#!/usr/bin/env bash
+# Copyright (C) 2022 Open Source Robotics Foundation
```
j-rivero, line 2: Maybe adding a `set -e` to stop on errors at the begging can help to detect hidden problems.

methylDragon:
> Seems like the script is in active development. If that is the case please mark it as Draft.
> 
> Before looking at the code, the script is placed under `jenkins-scripts/tools` but I believe that it does not have anything to do with Jenkins. We can create a new subdir in the root of the repo named something like `code-repo-scripts` to host the scripts designed to work with repository code. My intention is to move `source_changelog` into it but that will require a different PR.

I agree! I couldn't really find another place to put this into...

jenkins-scripts/tools/header_migration_script.sh
methylDragon: Good call! https://github.com/ignition-tooling/release-tools/pull/712/commits/cc67a0b0449189078bfeecf0199a29b328461477

chapulina:
>  We can create a new subdir in the root of the repo named something like code-repo-scripts to host the scripts designed to work with repository code. My intention is to move source_changelog into it but that will require a different PR.

I just saw this comment. On #677 I created the `source-repo-scripts` directory and moved `source_changelog` there.

methylDragon:
In the end it was still better to manually copy and move from the install space (too many edge cases), and it's unlikely we'll need to keep using this functionality, so I'm closing this :(

PR #714, opened 2022-04-27
Support for check_initialization_order in asan builds

j-rivero:
Designed to detect https://github.com/google/sanitizers/wiki/AddressSanitizerInitializationOrderFiasco. Currently no errors is found using this on ign-math since the problem should appear on "objects with static storage duration and fail if they are nontrivially constructible/destructible."

Tested DSL: [Build Status](https://build.osrfoundation.org/job/_dsl_ignition/1302/)
Tested build: <a href='https://build.osrfoundation.org/view/asan/job/ignition_math-ci_asan-ign-math6-focal-amd64/4/'></a>


j-rivero:
@azeey / @scpeters are we still interested in this change?

azeey:
Yeah, this looks good. Should we also enable `strict_init_order=true` since we should avoid any kind of static initialization dependencies?

j-rivero:
> Yeah, this looks good. Should we also enable `strict_init_order=true` since we should avoid any kind of static initialization dependencies?

I'll create a new PR for the `strict_init_order=true`. Merging this one first. Thanks!

PR #730, opened 2022-05-18
Auto detect PACKAGE_ALIAS to help with gz transition

j-rivero:
The PR implements the detection of new gz- names in source packages and modify the `PACKAGE_ALIAS` accordingly.

This should help with the transition of ign- packages to the new gz- names and make possible to work debbuilder without modifying the input parameters from release.py (nightly generation included).

The scape of `PACKAGE_ALIAS` variable needs to happen to get the new value generated in the script instead of hardcoding the value from input parameters.

### DIFF
--- a/jenkins-scripts/docker/lib/debbuild-base.bash
+++ b/jenkins-scripts/docker/lib/debbuild-base.bash
@@ -140,10 +140,16 @@ esac
 
 cd \${PACKAGE_RELEASE_DIR}
 
+# Helper for transition of ign to gz
+SRC_PACKAGE_NAME=\$(grep-dctrl -sSource -n  '' debian/control)
+if [[ \${SRC_PACKAGE_NAME} != \${SRC_PACKAGE_NAME/gz-} ]]; then
+  PACKAGE_ALIAS=\${SRC_PACKAGE_NAME}
+fi
+
 # [nightly] Adjust version in nightly mode
 if $NIGHTLY_MODE; then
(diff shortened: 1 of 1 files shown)


j-rivero:
Fix and new run  [Build Status](https://build.osrfoundation.org/job/ign-sensors7-debbuilder/385/)

jenkins-scripts/docker/lib/debbuild-base.bash
```diff
@@ -200,7 +207,7 @@ while (! \$update_done); do
 done
 
 # new versions of mk-build-deps > 2.21.1 left buildinfo and changes files in the code
-rm -f ${PACKAGE_ALIAS}-build-deps_*.{buildinfo,changes}
+rm -f \${PACKAGE_ALIAS}-build-deps_*.{buildinfo,changes}
```
scpeters, line 210: I don't understand the change on this line

j-rivero: All this code goes into a generated build.sh. There are two rounds for that script:
 1. Generation of the script: variable expanded in this first step are the ones coming from out of the own script, mainly input parameters of the Jenkins job.
 2. The script is being called with the variables without escaping being resolved at this point. If we leave `PACKAGE_ALIAS` here without escaping, the changes in this PR done to `PACKAGE_ALIAS` are not taking into account since it was already resolved. So we need to escape it in order to get the new value added in this PR some lines above.

Clarified or the doubt was different?

scpeters: thank you! some of the variable are not escaped (like line 218), which is why I asked

PR #732, opened 2022-05-19
Remove obsolete GITHUB_ORG and use new github orgs in debbuild

j-rivero:
Tested with new gz- packages [Build Status](https://build.osrfoundation.org/view/all/job/ign-tools2-debbuilder/92/)

### DIFF
--- a/jenkins-scripts/docker/lib/debbuild-base.bash
+++ b/jenkins-scripts/docker/lib/debbuild-base.bash
@@ -48,7 +48,7 @@ if ${NIGHTLY_MODE}; then
   if ${USE_REPO_DIRECTORY_FOR_NIGHTLY}; then
     mv ${WORKSPACE}/repo \$REAL_PACKAGE_NAME
   else
-    git clone https://github.com/gazebosim/\$REAL_PACKAGE_NAME -b ${NIGHTLY_SRC_BRANCH}
+    git clone https://github.com/${GITHUB_ORG}/\$REAL_PACKAGE_NAME -b ${NIGHTLY_SRC_BRANCH}
   fi
   PACKAGE_SRC_BUILD_DIR=\$REAL_PACKAGE_NAME
   cd \$REAL_PACKAGE_NAME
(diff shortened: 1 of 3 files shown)


jenkins-scripts/docker/lib/debbuild-base.bash
```diff
@@ -48,7 +48,7 @@ if ${NIGHTLY_MODE}; then
   if ${USE_REPO_DIRECTORY_FOR_NIGHTLY}; then
     mv ${WORKSPACE}/repo \$REAL_PACKAGE_NAME
   else
-    git clone https://github.com/${GITHUB_ORG}/\$REAL_PACKAGE_NAME -b ${NIGHTLY_SRC_BRANCH}
+    git clone https://github.com/gazebosim/\$REAL_PACKAGE_NAME -b ${NIGHTLY_SRC_BRANCH}
```
scpeters, line 51: gazebo classic is still on GitHub.com/osrf/gazebo

chapulina:
See https://github.com/gazebo-tooling/release-tools/pull/735 for an alternative that keeps osrf

chapulina:
Oh I guess now this and #735 are doing the same

chapulina (approved):

scpeters:
sorry I hadn't seen #735

chapulina:
> this isn't ready to merge because it doesn't handle the package renaming

Not sure what you mean, will this fail with the current package names?

scpeters:
> > this isn't ready to merge because it doesn't handle the package renaming
> 
> Not sure what you mean, will this fail with the current package names?

our code currently uses the old org name `ignitionrobotics` and the old repository names `ign-*`. This pull request updates the org name to `gazebosim` but retains the old repository names `ign-*`. I had assumed that this wouldn't work, but I did a quick check, and it looks like https://github.com/gazebosim/ign-cmake does actually forward properly, so I think my previous comment is incorrect

chapulina:
Ahh gotcha! I think this is because the repositories were moved incrementally:

1. https://github.com/ignitionrobotics/ign-cmake
2. https://github.com/gazebosim/ign-cmake
3. https://github.com/gazebosim/gz-cmake

PR #747, opened 2022-06-21
Explicit head branch when opening forward-port PR

chapulina:
I don't know exactly what conditions make this necessary, but I often run into a situation where the `gh` CLI complains like this:

`pull request create failed: GraphQL error: Head sha can't be blank...`

I noticed that passing an explicit head fixes the issue


chapulina:
I just had to use this again for https://github.com/gazebosim/gz-msgs/pull/264, so I'll merge it

PR #759, opened 2022-06-29
Add pytest to list of pip installed dependencies for sdformat

azeey:
As mentioned in https://github.com/gazebosim/sdformat/pull/1063, `pytest` is needed to generate junit xml files for python tests.

### DIFF
--- a/jenkins-scripts/sdformat-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/sdformat-default-devel-homebrew-amd64.bash
@@ -4,7 +4,7 @@
 [[ -L ${0} ]] && SCRIPT_DIR=$(readlink ${0}) || SCRIPT_DIR=${0}
 SCRIPT_DIR="${SCRIPT_DIR%/*}"
 
-PIP_PACKAGES_NEEDED="psutil"
+PIP_PACKAGES_NEEDED="psutil pytest"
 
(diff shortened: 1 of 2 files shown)


jenkins-scripts/sdformat-default-devel-homebrew-amd64.bash
```diff
@@ -4,7 +4,7 @@
 [[ -L ${0} ]] && SCRIPT_DIR=$(readlink ${0}) || SCRIPT_DIR=${0}
 SCRIPT_DIR="${SCRIPT_DIR%/*}"
 
-PIP_PACKAGES_NEEDED="psutil"
+PIP_PACKAGES_NEEDED="psutil pytest"
```
azeey, line 7: Now that I think about it, `gz-math` is going to need this as well. Should I put this in `project-default-devel-homebrew-amd64.bash` instead?

scpeters: that sounds good to me

azeey: Done in https://github.com/gazebo-tooling/release-tools/pull/759/commits/604cf628564277cdc5e7bb263b40b0f34048127d

j-rivero (approved):

PR #760, opened 2022-06-30
Use appropriate substitution invocation for COLCON_PACKAGE

methylDragon:
The wrong variable substitution directive is used for `COLCON_PACKAGE` for some of the windows jenkins scripts (`% %` does substitution at parse-time, while `! !` does substitution at runtime), causing variables to not get updated and for builds to fail.

This PR fixes that.


methylDragon:
[Build Status](https://build.osrfoundation.org/job/ign_sensors-pr-win/1027/)

PR #761, opened 2022-06-30
Install ffmpeg from vcpkg in gazebo builds.

nuclearsandwich:
This dependency was previously optional and seemingly omitted from Windows build farm builds. However it is now required based empirically on recently failing jobs.

Since ffmpeg is not a dependency that we've built and distributed via s3 we can try to install the version from vcpkg for now.

### DIFF
--- a/jenkins-scripts/lib/gazebo-base-windows.bat
+++ b/jenkins-scripts/lib/gazebo-base-windows.bat
@@ -22,6 +22,10 @@ call %win_lib% :remove_vcpkg_package qwt
 :: from vcpkg.
 call %win_lib% :install_vcpkg_package tinyxml2
 
+:: gazebo-sim needs libavutil (part of ffmpeg) which is not the win32 deps but
+:: it can be used from vcpkg.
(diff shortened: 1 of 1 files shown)


scpeters:
I assume this is in response to failures such as the following which fails due to a lack of the `av` component of ign-common / gz-common:

* https://build.osrfoundation.org/view/ign-garden/job/ign_gazebo-ci-win/185/

I believe `ffmpeg` is actually installed on these machines and that the problem is actually a failure to find the dependencies. See https://github.com/gazebosim/gz-common/issues/386 for more info.

scpeters:
see https://github.com/gazebosim/gz-cmake/pull/271

scpeters:
gz-sim windows builds are now unstable: https://github.com/gazebosim/gz-sim/pull/1578

Blast545:
The job is not failing anymore, so this PR can be closed. Steven! is OoO, as we opened this PR together, I'll take the freedom to close it. Thanks!

PR #775, opened 2022-07-28
merge forward script: ensure different args

scpeters:
I noticed a pull request created with the same branch name used for the source and destination. We can easily check for that case and return an error.


chapulina (approved):
Thanks! There's probably other sanity checks we can do too. Like it must be either `main` or start with `ign-` / `gz-`

PR #789, opened 2022-08-10
sdformat.dsl: rename internal variables

scpeters:
There are several `*_branches` variables that are actually sdformat package names (`sdformat6`, `sdformat9`, etc.) or `main`. This renames the variables to `*_versions` to try to clarify matters.


chapulina (approved):
Thanks, that's clearer

j-rivero (approved):

PR #753, opened 2022-06-24
ign -> gz : Release Tools Migrate Internal Vars

methylDragon:
Part of: https://github.com/gazebo-tooling/release-tools/pull/746

PR #792, opened 2022-08-16
Windows fixes for 753 (ign -> gz)

j-rivero:
@methylDragon while working on #753 I found that Windows builds [seems to fail ](https://build.osrfoundation.org/job/_test_ign_physics-ign-2-win/1/console)with problems in COLCON_PACKAGE var. We are chaging in #753 all COLCON_PACKAGE names to `gz-*`, something that is going to break old supported branches that expect `ignition-*`. I made 75877eec8c54725074b4cf996790ed14025477d8 to revert the logic.

The other patch here is removing the code in windows_library since the `build_workspace` is only called from the `jenkins-scripts/lib/colcon-default-devel-windows.bat` that already implements the logic as far as I can tell.

I've tested this branch in: https://build.osrfoundation.org/job/_test_ign_physics-ign-2-win/8/


methylDragon (approved):
LGTM

PR #794, opened 2022-08-19
Export Jenkins XML configuration diffs for DSL changed files in the CI

j-rivero:
Idea is easy: for the current DSL files modified we generate the real XML configuration in Jenkins in our CI currently to check syntax. With this PR we store the set of XML files generated, we use the master branch to get the current XML configurations in Jenkins and we run diff to export the changes between `master` and the PR so real differences appear clearly. The PR export two files: one with just the name of the config files changed and other with content changed.

Tested: https://github.com/j-rivero/release-tools/pull/1/files reporting https://github.com/j-rivero/release-tools/actions/runs/2886111942


j-rivero:
Need this for #799 . Merging.

Issue #412, opened 2021-03-04
Migration of CI Windows builds to use colcon


j-rivero:
Old system is using `configure.bat` files to help the compilation of the different Ignition libraries in CI by using the [generic-default-devel-windows.bat](https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/lib/generic-default-devel-windows.bat) script.

New approach uses [colcon-default-devel-windows.bat](https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/lib/colcon-default-devel-windows.bat), see [ign-gazebo](https://github.com/ignition-tooling/release-tools/blob/master/jenkins-scripts/ign_gazebo-default-devel-windows-amd64.bat) as an example.

Libraries pending of transition:

- [x] ign_cmake https://github.com/gazebo-tooling/release-tools/pull/812
- [x] ign_math https://github.com/ignition-tooling/release-tools/pull/602
- [x] ign_common https://github.com/ignition-tooling/release-tools/pull/511
- [x] ign_msgs https://github.com/ignition-tooling/release-tools/pull/536
- [x] ign_tools https://github.com/ignition-tooling/release-tools/pull/537
- [x] ign_fuel-tools
- [x] ign_transport https://github.com/ignition-tooling/release-tools/pull/411
- [x] ign_plugin https://github.com/ignition-tooling/release-tools/pull/495
- [x] sdformat https://github.com/ignition-tooling/release-tools/pull/365
- [ ] gazebo

scpeters:
it's necessary to update the GitHub branch settings to change which builds are required when the job names change

j-rivero:
Gazebo is not being tested in the buildfarm for Windows anymore, we can close this one.

PR #812, opened 2022-08-31
Port gz-cmake to use colcon on Windows

j-rivero:
This is the last gz- of #412. ign-cmake was a bit special since it adds a CMake parameters to the compilation thus I needed to implement that parameter passing in the form of `_COLCON_EXTRA_CMAKE_ARGS`.

If we want to keep the history of the CI jobs, we can rename them before we merge this PR. //cc @Blast545. This also affects t https://github.com/osrf/buildfarmer/pull/338.

Test:

 * New ign_cmake-ci-win [Build Status](https://build.osrfoundation.org/job/ign_cmake-ci-win/9/)
 * No use of cmake argument ign-plugin [Build Status](https://build.osrfoundation.org/job/ign_plugin-ci-win/31/)

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -92,7 +92,8 @@ gz_prerelease_pkgs    = [ 'placeholder' : [
                         ]
 // packages using colcon for windows compilation while migrating all them to
 // this solution
-gz_colcon_win         = [ 'common',
+gz_colcon_win         = [ 'cmake',
+                          'common',
                           'fuel-tools',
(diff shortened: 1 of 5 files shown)


scpeters (approved):

Blast545 (approved):
LGTM, if you can rename the job before merging I'd highly appreciate it :pray:

j-rivero:
Merged an renamed the -ci- jobs for the supported branches. I left the previous -pr- job in place so the current links in CI to it can work. New one should be in use from now on, that can duplicate the CI reports but I think that is better than losing the links.

PR #831, opened 2022-09-22
Use PR comments to display diffs in DSL->Jenkins changes

j-rivero:
Change the #794 to display the diffs between Jenkins XML config files generated for changes in the PR to use a comment in PR instead of exporting artifacts in Github Actions. Convenient to save 4-5 clicks, download and open different files when looking a PR.

The changes in the code are ugly since the handle of multiline output in combination with using ``` inside the github action brings many problems.

See it in action: https://github.com/j-rivero/release-tools/pull/3#issuecomment-1255403420

**Before merging:** it needs a token from an existing GitHub user with `public_repo` credentials in `USER_TOKEN`. The idea is to use @osrf-jenkins or  @gazebosim to do it.


scpeters (approved):
nice!

scpeters (commented):
LGTM, this is very cool

I really like the `diff` rendering

I'm not confident in my ability to review the javascript code used for the PR comment, so I'm not officially approving, but feel free to merge

PR #835, opened 2022-09-30
Fix terminal-dashboard to avoid confusions with arch=all packages

j-rivero:
The PR changes the approach used in the terminal dashboard since it is broken after the inclusion of the `ignition-*` transitional packages that arch-independent. When they are built for one arch, they appear in all the arches so the terminal dashboard wrongly assumes that there were packages built for the rest of arches.

New approach assumes that all the gz- libraries have a package named libgz-${LIB} except for the collection packages. This is now valid but somehow fragile. Real solution would imply a real parsing and add logic which is quite an effort much bigger.

PR also takes i386 out from new collections starting with Debian.

### DIFF
--- a/terminal-dashboard/table.bash
+++ b/terminal-dashboard/table.bash
@@ -32,12 +32,24 @@ ARCHS=( "amd64")
 DISTROS=( "ubuntu" )
 # No nightlies or pre-releases for arm
 if [[ $PACKAGE_REPO == "stable" ]]; then
-  ARCHS+=( "i386" "arm64" "armhf")
+  if [[ $COLLECTION == "citadel" || $COLLECTION == "fortress" ]]; then
+    ARCHS+=( "i386" )
+  fi
+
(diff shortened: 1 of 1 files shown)


scpeters:
> New approach assumes that all the gz- libraries have a package named libgz-${LIB} except for the collection packages. This is now valid but somehow fragile. 

I think there is no `libgz-cmake3`; it only has a `-dev` package

j-rivero:
> I think there is no `libgz-cmake3`; it only has a `-dev` package

You are right. I have changed the logic and documentation in [674bfe9](https://github.com/gazebo-tooling/release-tools/pull/835/commits/674bfe9f79bcc12468097dcc3edf65b544152c98) to use libgzfoo or libgzfoo-dev since both are arch independent and should make the signal of a successful build.

j-rivero:
Working fine on my tests with Citadel and Garden.

PR #858, opened 2022-11-18
Better support for "none" in UPLOAD_TO_REPO param in Jenkins debbuilders

j-rivero:
For testing proposes when we test the -release repo changes or other packaging stuff, this PR improves the "unofficial" support for `none` value in the `UPLOAD_TO_REPO` parameter:

 * DSL code change: ddede4bcc609a4c28b5c2eba9be8337dd6bdfe30 Avoid the call to `repository_uploader` if none is passed in `UPLOAD_TO_REPO`.
 * Changes in repsitory_uploader script: in the case that the 'none' value is found in the `UPLOAD_TO_REPO` parameter do not execute any upload code and finish successfully. bad3c51f177f2e7a81efbc00c3cda2b587b3877d

Note: the tabs were very broken in the code, better see this PR [with ?w=1](https://github.com/gazebo-tooling/release-tools/pull/858/files?w=1)

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFLinuxBuildPkg.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFLinuxBuildPkg.groovy
@@ -99,21 +99,26 @@ class OSRFLinuxBuildPkg
             failBuildOnError(false)
         }
 
-        conditionalAction {
-          condition {
-            not {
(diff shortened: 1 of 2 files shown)


azeey (approved):
This is great Jose! I've been using a hack: rebuilding an existing job and keeping the same version number so it doesn't upload. I think it's great we officially support the `none` upload option. I would encourage to always use this when testing PRs.

PR #874, opened 2023-01-26
Use system s3cmd instead of using 1.5.0

j-rivero:
Using old s3cmd is looking for python instead of python3. Remove and use the system provided one with new provision.


j-rivero:
Tested here [Build Status](https://build.osrfoundation.org/job/repository_uploader_packages/35847/)

PR #883, opened 2023-03-06
brew without sudo: part 2

scpeters:
Add group write permissions without sudo. I'm not sure if this will work or not. Testing below:

### Simple test:

* [Build Status](https://build.osrfoundation.org/view/ign-garden/job/ignition_utils-ci-gz-utils2-homebrew-amd64/22/) https://build.osrfoundation.org/view/ign-garden/job/ignition_utils-ci-gz-utils2-homebrew-amd64/22/


j-rivero (approved):
> Add group write permissions without sudo. I'm not sure if this will work or not. Testing below:

Umm seems to work. I can not find a place where we are adding stuff to `$HOMEBREW_CELLAR` with other permissions different than the current user.

scpeters:
let's try it and see if we get any problems

PR #884, opened 2023-03-07
Update the Garden collection/view on Jenkins

azeey:
I removed some redundant entries (`gz_*-install-pkg-focal-amd64`), use the `gz-sim7` job instead of `main` for gz-sim, and moved sdformat jobs to be grouped with similar jobs.

Questions for @j-rivero / @scpeters :

- I noticed that `ignition_sim7-install_bottle-homebrew-amd64` is missing. Does this exist under a different name?
- Also, I noticed the naming pattern changed for gz-sim, e.g, instead of `ign_gazebo-gz-7-win`, we now use `ign_gazebo-gz-sim7-win`. Not sure if that's intended.

PR #885, opened 2023-03-08
Fix generation of sim7 bottle

j-rivero:
Following up comment in https://github.com/gazebo-tooling/release-tools/pull/884#issue-1614016367

### DIFF
--- a/jenkins-scripts/dsl/ignition.dsl
+++ b/jenkins-scripts/dsl/ignition.dsl
@@ -695,8 +695,8 @@ gz_software.each { gz_sw ->
   }
 
   // 3. install jobs to test bottles
-  supported_install_pkg_branches(software_name).each { major_version, supported_distros ->
-    def install_default_job = job("ignition_${software_name}${major_version}-install_bottle-homebrew-amd64")
+  supported_install_pkg_branches(gz_sw).each { major_version, supported_distros ->
+    def install_default_job = job("ignition_${gz_sw}${major_version}-install_bottle-homebrew-amd64")
     OSRFBrewInstall.create(install_default_job)
 
     install_default_job.with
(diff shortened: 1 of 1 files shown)


j-rivero:
From CI job:
```diff
Only in /tmp/pr_xml_configuration: ignition_sim7-install_bottle-homebrew-amd64.xml
```
The new configuration is generating the sim7 install bottle now.

scpeters:
this job isn't working because there is only a `gz-sim7` formula but no `ignition-sim7` alias

* [Build Status](https://build.osrfoundation.org/job/ignition_sim7-install_bottle-homebrew-amd64/2/) https://build.osrfoundation.org/job/ignition_sim7-install_bottle-homebrew-amd64/2/

~~~
+ brew install --include-test ignition-sim7
Warning: No available formula with the name "ignition-sim7". Did you mean ignition-msgs5, ignition-msgs1, ignition-msgs0, ignition-msgs8, ignition-msgs2, ignition-gui6 or ignition-gui3?
~~~

* https://github.com/osrf/homebrew-simulation/blob/master/Formula/gz-sim7.rb

The `ignition-gazebo*` formula have corresponding `gz-sim` aliases.

* https://github.com/osrf/homebrew-simulation/blob/master/Formula/ignition-gazebo3.rb
* https://github.com/osrf/homebrew-simulation/blob/master/Aliases/gz-sim3
* https://github.com/osrf/homebrew-simulation/blob/master/Formula/ignition-gazebo6.rb
* https://github.com/osrf/homebrew-simulation/blob/master/Aliases/gz-sim6

I suppose the simplest workaround would be to add an `ignition-sim7` alias to `gz-sim7`. It would be cleaner to avoid this so users don't start to depend on it, but we can do it if needed.

j-rivero:
> this job isn't working because there is only a `gz-sim7` formula but no `ignition-sim7` alias

quick fix https://github.com/gazebo-tooling/release-tools/pull/887

PR #893, opened 2023-03-27
Avoid docker cache for SOURCE_DEFINED_DEPS

j-rivero:
Caching gz packages is generating problems when gzdev changes the configuration of project to install different packages.osrfoundation.org repositories (stable, nightly or prerelease), i.e: https://github.com/gazebo-tooling/gzdev/pull/64.

When using a new repository that expects to have newer versions (i.e: prerelease on top of stable) this is perfectly fine and the release scripts will work nicely. But when that change is reverted and the build needs to remove a repository with newer versions installed, that is problematic since the packages are already in docker cache installed.

This PR removes the caching of `SOURCE_DEFINED_DEPS` to leave mostly base system packages cached. It also removes the monthly invalidation since seems to me to have a little use and simplifies the code.

### DIFF
--- a/jenkins-scripts/docker/lib/docker_generate_dockerfile.bash
+++ b/jenkins-scripts/docker/lib/docker_generate_dockerfile.bash
@@ -295,7 +295,7 @@ cat >> Dockerfile << DELIM_DOCKER31
 # to run apt-get update again
 RUN (apt-get update || (rm -rf /var/lib/apt/lists/* && apt-get update)) \
  && apt-get dist-upgrade -y \
- && apt-get install -y ${SOURCE_DEFINED_DEPS}
+ && apt-get install -y ${SOURCE_DEFINED_DEPS} \
  && apt-get clean
 
 # Map the workspace into the container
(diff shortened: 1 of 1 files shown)


azeey:
I thought the issue was that the nightly or prerelease repos are enabled in the cache, not that there are packages installed. I'm not sure if this fixes that since the `dockerfile_install_gzdev_repos` call happens earlier in the file (L207) than the changes in this PR. I also see `dockerfile_install_gzdev_repos` on L291, but unless `gzdev` overwrites all the gazebo related files in `/etc/apt/sources.list.d`, we'd still have caching issues. Don't we need to the `gzdev.py repository enable` call in `build.sh` to avoid any caching?

azeey:
Any updates on this @j-rivero ?

j-rivero:
> Any updates on this @j-rivero ?

I think that you are right here. I spotted the problem with the packages being installed from an unstable repository when stable builds run but did not check the status of the installed repos. Moving this to `build.sh` could be a better approach, indeed. Cancelling this and open a new PR for it.

PR #902, opened 2023-04-04
ABI script: get correct header paths

### LINKED ISSUE #890: ABI checker script has trouble identifying header path
The [generic-abi-base.bash](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/docker/lib/generic-abi-base.bash) script uses brittle logic to identify the [DEST_DIR](https://github.com/gazebo-tooling/release-tools/blob/2ae15207759fb377539fe20fbda9c492541842f9/jenkins-scripts/docker/lib/generic-abi-base.bash#L71) and [SRC_DIR](https://github.com/gazebo-tooling/release-tools/blob/2ae15207759fb377539fe20fbda9c492541842f9/jenkins-scripts/docker/lib/generic-abi-base.bash#L92). Specifically, it looks for a folder containing the project name followed by a `-` character (`${ABI_JOB_SOFTWARE_NAME}-*`). This only currently works for SDFormat, which installs headers to a folder like the following:

* `<prefix>/include/sdformat-9.10`

It used to work for `gazebo`, which uses a `gazebo-11` style, but the `ABI_JOB_SOFTWARE_NAME` recently changed to `gazebo-classic` to match the repository name, so it no longer matches and the ABI jobs are broken (see https://github.com/gazebosim/gazebo-classic/issues/3292).

This logic also doesn't work for `ignition-*` or `gz-*` libraries, which use the following install pattern:

* `<prefix>/include/ignition/math6/`
* `<prefix>/include/gz/math7/`

scpeters:
Fixes #890.

The ABI script currently assumes that the headers are installed in a subfolder that
starts with the `ABI_JOB_HEADER_PREFIX` variable and a `-`, which only works for `sdformat-9` but requires adjustment for most other packages. This adds an `ABI_JOB_HEADER_PREFIX` variable to help locate the header folder.

### DIFF
--- a/jenkins-scripts/docker/gazebo-abichecker.bash
+++ b/jenkins-scripts/docker/gazebo-abichecker.bash
@@ -17,6 +17,7 @@ fi
 . ${SCRIPT_DIR}/lib/_gazebo_version_hook.bash
 
 export ABI_JOB_SOFTWARE_NAME="gazebo-classic"
+export ABI_JOB_HEADER_PREFIX="gazebo-*"
 export ABI_JOB_REPOS="stable"
 export ABI_JOB_PKG_DEPENDENCIES_VAR_NAME="GAZEBO_BASE_DEPENDENCIES"
 export ABI_JOB_IGNORE_HEADERS="gazebo/GIMPACT gazebo/opcode gazebo/test"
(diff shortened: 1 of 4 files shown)


scpeters:
this needs an adjustment for `ign-gazebo` / `gz-sim` depending on the major version

j-rivero (approved):
I don't see any problem.

PR #905, opened 2023-04-11
DSL: ignore nodes labeled as test-instance

j-rivero:
Support to define a node label called `test-instance` that will be ignored by all the works.

The implementation adds a new helper function to Globals to supply the logic to ignore the test label.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/Globals.groovy
+++ b/jenkins-scripts/dsl/_configs_/Globals.groovy
@@ -119,4 +119,8 @@ class Globals
    static String get_ros2_development_distro() {
      return 'rolling'
    }
+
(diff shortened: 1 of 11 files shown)


nuclearsandwich (approved):
Looks great! The use of a global helper is going to make maintaining labels that much easier!

When setting up test jobs we can duplicate a job run and remove the block on test instances.

nuclearsandwich (approved):
Looks great! The use of a global helper is going to make maintaining labels that much easier!

When setting up test jobs we can duplicate a job run and remove the block on test instances.

j-rivero:
Changes will land tomorrow when DSL generation runs nightly. If you need them before, please let me know.

Issue #910, opened 2023-04-17
Multiple MSVC compilers in an image make VCPKG to use the latest by default


j-rivero:
If several compiler are installed in the same Windows system, even if the `vcvarsall.bat` is called when one of the versions that is not the latest one, vcpkg always prefer the latest version unless `VCPKG_PLATFORM_TOOLSET` is configured inside the triplets files.

More information https://github.com/Microsoft/vcpkg/issues/1234
TOOLSETs: https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_platform_toolset

PR #911, opened 2023-04-17
Explicit support for MSVC_VERSION_REQUIRED in release-tools

j-rivero:
Initial work to support multiple compilers on Windows and make the script to explicitly declared which one to use.

The 2022 MSVC support is left commented until #910 is fixed . Leave the code enabled will introduce a hidden bug since vcpkg will be using 2022 is available when the existing distributions are using 2019.

Tested: [Build Status](https://build.osrfoundation.org/job/ign_utils-pr-win/301/)


j-rivero:
This is going to change with the Conda transition. Closing.

Issue #912, opened 2023-04-18
`nvml error: driver/library version mismatch` in Linux GPU agents


Crola1702:
Sometimes, Linux GPU Agents make build fail early because of unattended-upgrades.

Reference build: https://build.osrfoundation.org/job/ignition_gazebo-ci-gz-sim7-focal-amd64/91/console

Log output:
```
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #1: error running hook: exit status 1, stdout: , stderr: Auto-detected mode as 'legacy'
nvidia-container-cli: initialization error: nvml error: driver/library version mismatch: unknown.
Build step 'Execute shell' marked build as failure
```

The solution for now is rebooting the agent each time it happens (check this [StackOverflow answer](https://stackoverflow.com/a/43023000/19846687)), but we should automate the response to this problem via:
1. catch this error in script and send a sigil to the Jenkins post-build runner which
2. marks the Jenkins host offline
3. re-queues the job
4. forces a reboot of the Jenkins agent

j-rivero:
>  Linux GPU Agents make build fail early because of unattended-upgrades.

I'm assuming that we prefer to have `unattended-upgrades` running on all the packages that trying to pin/freeze the affected packages so we don't have the problem until we decide to update the image for a good reason.

> 1. catch this error in script and send a sigil to the Jenkins post-build runner which

I've being using the ["Failure Cause Management" plugin](https://build.osrfoundation.org/failure-cause-management/) (configuration[ is stored in chef)](https://github.com/osrf/chef-osrf/blob/latest/cookbooks/osrfbuild/files/default/build-failure-analyzer.xml) for a while to notify a message in the main build page when a regexp is being found but a part of this I don't think that we can do much more with that plugin.

> 2. marks the Jenkins host offline
> 3. re-queues the job

The ros_buidlfarm repo hosts groovy scripts that can serve you to create a custom solution for this. For example [this one checks for a regexp in the log](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/snippet/builder_system-groovy_extract-warnings.xml.em#L14-L26) or this one [disables a node and log the problem](https://github.com/ros-infrastructure/ros_buildfarm/blob/master/ros_buildfarm/templates/snippet/builder_system-groovy_check-free-disk-space.xml.em#L32-L49).

claraberendsen:
[...]

PR #915, opened 2023-04-25
nvidia mismatch recovery

claraberendsen:
**DESCRIPTION**

Since  moving to cloud instances for the `nvidia agents` we have encountered issues with versioning resulting in `nvml version mismatch errors` (See [issue](https://github.com/gazebo-tooling/release-tools/issues/912)). This is an unrecoverable error for the builds, and requires manual steps to bring back the agent to a funtional state. 
This PR automates those steps as postbuild actions. If the build failed due to the Nvidia error, we do the following steps: 
1. Remove all labels from current agent and add a `recovery-process` label to indicate that that agent is performing recovery actions, preventing it from taking on any other build.
2. Requeue the job with same parameters with a delay of 70s to give time for the next step. 
3. Schedule a system restart on 1 minute (the delay is needed here so that the postbuild action finishes correctly)

**TESTS**

- [x] Post build actions only run on failed builds [Build Status](https://build.osrfoundation.org/job/_test_job_from_dsl/41/)
- [x]  Post build actions only run on nvidia agents [Build Status](https://build.osrfoundation.org/job/_test_job_from_dsl/46/)
- [x]  Steps run only if `regex` match [Build Status](https://build.osrfoundation.org/job/_test_job_from_dsl/61/)
- [x]  Adds`dummy` not actionable labels to agent to not build jobs between current and restart

- [x] Agent is put back online with correct labels after restart.

- [x] Rebuilds job with same parameters
        Manual rebuild: https://build.osrfoundation.org/job/_test_job_from_dsl/80/
        Automatic rebuild after recovery process: https://build.osrfoundation.org/job/_test_job_from_dsl/81/

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
@@ -42,6 +42,13 @@ class OSRFUNIXBase extends OSRFBase
              """.stripIndent())
       }
       publishers {
+        configure { project -> 
+          project / publishers / 'com.chikli.hudson.plugin.naginator.NaginatorPublisher' {
+            regexpForRerun(this.regexNvidiaLabel)
+            checkRegexp(true)
+          }
+
+        }
         postBuildScripts {
           steps{
             conditionalSteps {
@@ -85,10 +92,6 @@ class OSRFUNIXBase extends OSRFBase
         retryBuild {
           fixedDelay(70)
           retryLimit(1)
-          configure { 
-            regexpForRerun(this.regexNvidiaLabel)
-            checkRegexp(true)
-          }
         }
       }
     }
--- a/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
@@ -39,17 +39,6 @@ class OSRFUNIXBase extends OSRFBase
              """.stripIndent())
       }
       publishers {
-        // Manual insertion of xml for Naginator plugin because of this issue https://issues.jenkins.io/browse/JENKINS-66458
-        configure { project -> 
-          project / publishers / 'com.chikli.hudson.plugin.naginator.NaginatorPublisher' {
-            regexpForRerun("nvml error: driver/library version mismatch")
-            checkRegexp(true)
-            maxSchedule(1)
-            delay(class: 'com.chikli.hudson.plugin.naginator.FixedDelay') {
-              delay(70)
-            }
-          }
-        }
         postBuildScripts {
           steps{
             conditionalSteps {
@@ -89,6 +78,17 @@ class OSRFUNIXBase extends OSRFBase
           }
           onlyIfBuildSucceeds(false)
           onlyIfBuildFails(true)
+        }
+        // Manual insertion of xml for Naginator plugin because of this issue https://issues.jenkins.io/browse/JENKINS-66458
+        configure { project -> 
+          project / publishers / 'com.chikli.hudson.plugin.naginator.NaginatorPublisher' {
+            regexpForRerun("nvml error: driver/library version mismatch")
+            checkRegexp(true)
+            maxSchedule(1)
+            delay(class: 'com.chikli.hudson.plugin.naginator.FixedDelay') {
+              delay(70)
+            }
+          }
         }
       }
     }
--- a/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
@@ -10,10 +10,10 @@ import javaposse.jobdsl.dsl.Job
      - allow concurrent builds
      - bash: RTOOLS checkout
 */
-class OSRFUNIXBase extends OSRFBase 
+class OSRFUNIXBase extends OSRFBase
 {
-  static void create(Job job) 
-  { 
+  static void create(Job job)
+  {
     OSRFBase.create(job)
 
     job.with
@@ -96,5 +96,4 @@ class OSRFUNIXBase extends OSRFBase
         
     }
   }
-
 }
--- a/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
@@ -55,7 +55,7 @@ class OSRFUNIXBase extends OSRFBase
 
                         println("# BEGIN SECTION: NVIDIA MISMATCH RECOVERY")
                         if (!(build.getLog(1000) =~ "nvml error: driver/library version mismatch")) {
-                          println("Build failed for other reason than NVIDIA error - Not performing any more steps")
+                          println(" NVIDIA driver/library version mismatch not detected in the log - Not performing any recovery automatic recovery step")
                           return 0;
                         } else {
                           try {
--- a/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
@@ -38,62 +38,61 @@ class OSRFUNIXBase extends OSRFBase
              git clone https://github.com/gazebo-tooling/release-tools scripts -b \$RTOOLS_BRANCH
              """.stripIndent())
       }
-          publishers {
-            postBuildScripts {
-              steps{
-              conditionalSteps {
-                  condition {
-                      expression('(.)*gpu-nvidia(.)*','${NODE_LABELS}')
-                  }
-                  steps {
-                     systemGroovyCommand('''\
-                        import hudson.model.Cause.UpstreamCause;
-                        import hudson.model.*;
-                    
-                        def node = build.getBuiltOn()
-                        def old_labels = node.getLabelString()
+      publishers {
+        postBuildScripts {
+          steps{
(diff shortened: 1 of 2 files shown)


claraberendsen:
Build that succeeded after restart 
[Build Status](https://build.osrfoundation.org/job/_test_ignition_gazebo-ci-gz-sim7-focal-amd64/7/)
Build that restarted the agent
[Build Status](https://build.osrfoundation.org/job/_test_job_from_dsl/33/)

**WIP**
- Job fails with agent disconnection (will add a sleep so the build can finish). 
- Rebuild job, code commented. Need to test it further.

claraberendsen:
For documentation purposes to test the log output, I'm manually adding to the test job `_test_job_from_dsl` a build step that echos the text I'm looking for. 
Will do a PR to add this as a param, so we can have a test job that we can inject any necessary text in the logs. It is also possible to add another param, to force a STATUS for the build (e.g. force the build to fail).

claraberendsen:
@j-rivero would appreciate a review!

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -38,6 +38,62 @@ class OSRFUNIXBase extends OSRFBase
              git clone https://github.com/gazebo-tooling/release-tools scripts -b \$RTOOLS_BRANCH
              """.stripIndent())
       }
+          publishers {
+            postBuildScripts {
+              steps{
+              conditionalSteps {
+                  condition {
+                      expression('(.)*gpu-nvidia(.)*','${NODE_LABELS}')
```
j-rivero, line 46: Can not find the official documentation for the `NODE_LABELS` variable. If it works as expected, it is probably a white space separated list of labels. If that is the case, we can refine the regexp to match exactly `gpu-nvidia` and not other hypothetical variants line `no-gpu-nvidia`.

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -38,6 +38,62 @@ class OSRFUNIXBase extends OSRFBase
...
+                     systemGroovyCommand('''\
+                        import hudson.model.Cause.UpstreamCause;
+                        import hudson.model.*;
+                    
+                        def node = build.getBuiltOn()
+                        def old_labels = node.getLabelString()
+
+                        println("# BEGIN SECTION: NVIDIA MISMATCH RECOVERY")
+                        if (!(build.getLog(1000) =~ "nvml error: driver/library version mismatch")) {
+                          println("Build failed for other reason than NVIDIA error - Not performing any more steps")
```
j-rivero, line 58: nitpick: a bit more accurate on what we are really checking:
```suggestion
                          println(" NVIDIA driver/library version mismatch not detected in the log - Not performing any recovery automatic recovery step")
```

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -38,6 +38,62 @@ class OSRFUNIXBase extends OSRFBase
...
+                          return 0;
+                        } else {
+                          try {
+                            println("Removing labels and adding 'recovery-process' label to node")
```
j-rivero, line 62: Here we have detected the nvidia problem. May be can add a message before this one saying something like:
```groovy
println(" PROBLEM: NVIDIA driver/library version mismatch was detected in the log. Try to automatically resolve it:")
```

j-rivero (commented):
While reviewing the PR, in the section of passing the parameters to rebuild the failed build I was looking into the [naginator plugin](https://plugins.jenkins.io/naginator/) that we have in the buildfarm which usually does a really good work in not successful builds just pressing "Retry" on the GUI. The plugin claims to support: "Only rebuild the job if the build's log output contains a given regular expression" which seems to me like it could be the same thing that we are implementing with custom code here. The GUI is:

image

The custom implementation here has some advantages like restricting it to only relevant gpu nodes on Linux but the use of a plugin can be more reliable that maintaining and debugging custom groovy code. Before I continue with the review, what do you think @claraberendsen , that we can keep going with this PR or move it to use the retry plugin features?

nuclearsandwich:
> Remove all labels from current agent and add a `recovery-process` label to indicate that that agent is performing recovery actions, preventing it from taking on any other build.

As this PR proceeds I want to raise an alternative tactic we could take and why I think the current action is the preferred one. Instead of dropping all other labels, we could add the `recovery-process` label and maintain the others. However, doing that without accounting for it everywhere would mean that singleton labels applied to the node being recovered would still show up as present while the node is no longer running jobs. I was slightly fearful of destructive edits to labels but I think it's actually the best thing to do with autorecovery scripts. 👍🏼

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
claraberendsen: The NODE_LABELS variable is documented in https://github.com/jenkinsci/job-dsl-plugin/wiki/User-Power-Moves as an option. And yes it's a white space separated list of variables since I think  Jenkins treats variables as that and matches against it. I will tune the regex so it matches only gpu-nvidia.

claraberendsen:
> ere has some advantages like restricting it to only relevant gpu nodes on Linux but the use of a plugin can be more reliable that maintaining and debugging custom groovy code. Before I continue with the review, what do you think @claraberendsen , that we can keep going with this PR or move it to use the retry plugin features?



> While reviewing the PR, in the section of passing the parameters to rebuild the failed build I was looking into the [naginator plugin](https://plugins.jenkins.io/naginator/) that we have in the buildfarm which usually does a really good work in not successful builds just pressing "Retry" on the GUI. The plugin claims to support: "Only rebuild the job if the build's log output contains a given regular expression" which seems to me like it could be the same thing that we are implementing with custom code here. The GUI is:
> 
> image
> 
> The custom implementation here has some advantages like restricting it to only relevant gpu nodes on Linux but the use of a plugin can be more reliable that maintaining and debugging custom groovy code. Before I continue with the review, what do you think @claraberendsen , that we can keep going with this PR or move it to use the retry plugin features?

I tried the retry plugin and I found this particular issue: 
- I could not find the dsl docs on the only re-run if expression is found part. [This](https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.retryBuild) is the docs I was using as reference. 

If you know of a way to use that conditional on dsl I would prefer that strategy over the current one for maintainability. On the custom run only on gpu nodes, with the naginator plugin even though we won't have complete certainty that it will run only on gpu-nvidia nodes, we are not expecting to encounter this log in a normal workflow in any other node. The only case I could see right now is if someone with intent adds that log to a build it would force the rebuild of the job and we won't be able to stop that.

claraberendsen:
**Tests after changes**
- [] Post build actions only run on failed builds 
- []  Post build actions only run on nvidia agents
- []  Steps run only if `regex` match
- [] Rebuilds job with same parameters

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -38,6 +38,59 @@ class OSRFUNIXBase extends OSRFBase
...
+                      println("ERROR - CANNOT PERFORM RECOVERY ACTIONS FOR NVIDIA ERROR")
+                      println("Restoring to previous state")
+                      node.setLabelString(old_labels)
+                      throw ex
+                    }
+                  }
+                  println("# END SECTION: NVIDIA MISMATCH RECOVERY")
+                  '''.stripIndent()
+                )
+                shell("""sudo shutdown -r +1""")
```
j-rivero, line 86: Can we double check that this shutdown step is only being run when we detect the nvidia problem and not unconditionally as part of the step sequence?

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -38,6 +38,59 @@ class OSRFUNIXBase extends OSRFBase
              git clone https://github.com/gazebo-tooling/release-tools scripts -b \$RTOOLS_BRANCH
              """.stripIndent())
       }
+      publishers {
+        // Manual insertion of xml for Naginator plugin because of this issue https://issues.jenkins.io/browse/JENKINS-66458
+        configure { project ->
```
j-rivero, line 43: We probably want to move the retry/naginator configure block after the `postBuildScripts` so the remove of the labels in the node happens before we launch the build back.

claraberendsen: Will run a test job once the debbuilders finish running. The logic is that since that shutdown is on a step block, whenever we don't find the regex the previous step fails and therefore the shutdown is not run.

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
claraberendsen: Will run a test job once the debbuilders finish running. The logic is that since that shutdown is on a step block, whenever we don't find the regex the previous step fails and therefore the shutdown is not run.

claraberendsen: Now it's not running when no nvidia problem detected! [Build test](https://build.osrfoundation.org/job/_test_job_from_dsl/93/)

j-rivero (approved):
Good to go. Great work @claraberendsen !

PR #916, opened 2023-04-28
Make display detection more robust.

### LINKED ISSUE #906: `DISPLAY` environment variable race condition when checking GPU support
Reference build: https://build.osrfoundation.org/job/_test_ignition_gazebo-ci-gz-sim7-focal-amd64/1/ (Test job)

Log output: 

```
++++ echo 'GPU support needed by the script but DISPLAY var is empty'
GPU support needed by the script but DISPLAY var is empty
++++ sudo service lightdm restart
++++ export_display_variable
+++++ ls /tmp/.X11-unix/
+++++ sed -e 's@^X@:@'
++++ [[ '' == '' ]]
++++ echo 'Imposible to get DISPLAY variable. Check your system'
Imposible to get DISPLAY variable. Check your system
++++ exit 1
Build step 'Execute shell' marked build as failure
```

When debugging jobs in the new cloud linux agents. We faced an issue with the `DISPLAY` environment variable. It seems it's caused when we restart lightdm service using `sudo service lightdm restart` and then running `export_display_variable`.

In a conversation with @nuclearsandwich he mentioned that this issue could be because of a race condition in the lightdm restart scripts mainly because it needs to wait an extra few seconds for X11 to come up. This could make the first build of an agent fail sometimes.

Reference lines:
* [sudo sercive lightdm restart](https://github.com/gazebo-tooling/release-tools/blob/956ef284c52bd6261b46ac7c37f25071b663d70d/jenkins-scripts/lib/check_graphic_card.bash#L82)
* [export_display_variable](https://github.com/gazebo-tooling/release-tools/blob/956ef284c52bd6261b46ac7c37f25071b663d70d/jenkins-scripts/lib/check_graphic_card.bash#L84)

nuclearsandwich:
Attempting to improve the display detection for the Linux GPU agents to resolve #906.

### DIFF
--- a/jenkins-scripts/lib/check_graphic_card.bash
+++ b/jenkins-scripts/lib/check_graphic_card.bash
@@ -12,7 +12,7 @@ export_display_variable()
         # Strip the path and leading X from the X11 socket
         # but check that the resulting string is numeric and
         # non-empty before exporting.
-        DISPLAY=$(basename $i | sed -n -E 's:^X([0-9]+):\1:')
+        DISPLAY=$(basename $i | sed -n -E 's:^X([0-9]+):\1:p')
         if [ -n $DISPLAY ]; then
           export DISPLAY
         fi
--- a/jenkins-scripts/lib/check_graphic_card.bash
+++ b/jenkins-scripts/lib/check_graphic_card.bash
@@ -3,16 +3,19 @@ GRAPHIC_CARD_PKG=""
 
 export_display_variable()
 {
-    # Hack to found the current display (if available) two steps:
-    # Check for /tmp/.X11-unix/ socket and check if the process is running
-    for i in $(ls /tmp/.X11-unix/ | sed -e 's@^X@:@')
+    # Check for an active X11 display socket.
+    for i in $(find /tmp/.X11-unix -type s)
     do
-      # grep can fail so let's disable the fail or error during its call
-      set +e
-      ps aux | grep bin/X.*$i | grep -v grep
-      set -e
-      if [ $? -eq 0 ] ; then
-	export DISPLAY=$i
+      # If a process is running with the open socket then
+      # lsof will exit successfully.
+      if lsof -Fp $i ; then
+        # Strip the path and leading X from the X11 socket
+        # but check that the resulting string is numeric and
+        # non-empty before exporting.
+        DISPLAY=$(basename $i | sed -n -E 's:^X([0-9]+):\1:')
+        if [ -n $DISPLAY ]; then
+          export DISPLAY
+        fi
       fi
     done
 }
--- a/jenkins-scripts/lib/check_graphic_card.bash
+++ b/jenkins-scripts/lib/check_graphic_card.bash
@@ -13,7 +13,7 @@ export_display_variable()
         # but check that the resulting string is numeric and
         # non-empty before exporting.
         DISPLAY=:$(basename $i | sed -n -E 's/^X([0-9]+)/:\1/p')
-        if [ -n $DISPLAY ]; then
+        if [ -n "$DISPLAY" ]; then
           export DISPLAY
         fi
       fi
--- a/jenkins-scripts/lib/check_graphic_card.bash
+++ b/jenkins-scripts/lib/check_graphic_card.bash
@@ -83,6 +83,8 @@ if $GPU_SUPPORT_NEEDED; then
       echo "GPU support needed by the script but DISPLAY var is empty"
       # Try to restart lightdm. It should stop the script in the case of failure
       sudo service lightdm restart
+      # Wait for lightdm service to restart X11.
+      sleep 5
       # Second try to get display variable
       export_display_variable
       if [[ ${DISPLAY} == "" ]]; then
--- a/jenkins-scripts/lib/check_graphic_card.bash
+++ b/jenkins-scripts/lib/check_graphic_card.bash
@@ -82,7 +82,7 @@ if $GPU_SUPPORT_NEEDED; then
     if [[ ${DISPLAY} == "" ]]; then
       echo "GPU support needed by the script but DISPLAY var is empty"
       # Try to restart lightdm. It should stop the script in the case of failure
-      sudo service lightdm restart
+      sudo systemctl restart lightdm
       # Wait for lightdm service to restart X11.
       sleep 5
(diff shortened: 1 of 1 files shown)


jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -3,16 +3,19 @@ GRAPHIC_CARD_PKG=""
 
 export_display_variable()
 {
-    # Hack to found the current display (if available) two steps:
-    # Check for /tmp/.X11-unix/ socket and check if the process is running
-    for i in $(ls /tmp/.X11-unix/ | sed -e 's@^X@:@')
+    # Check for an active X11 display socket.
+    for i in $(find /tmp/.X11-unix -type s)
```
nuclearsandwich, line 7: Using `find` instead of `ls` makes it possible to print out specifically sockets within the directory rather than all files.

jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -3,16 +3,19 @@ GRAPHIC_CARD_PKG=""
...
     do
-      # grep can fail so let's disable the fail or error during its call
-      set +e
-      ps aux | grep bin/X.*$i | grep -v grep
-      set -e
-      if [ $? -eq 0 ] ; then
-	export DISPLAY=$i
+      # If a process is running with the open socket then
+      # lsof will exit successfully.
+      if lsof -Fp $i ; then
```
nuclearsandwich, line 11: With lsof we can "filter" socket files by whether or not an active process has each open rather than pattern matching all running processes to string match on a likely candidate.

Using it in the conditional directly also saves us having to bump around `set +e/-e`.

jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -3,16 +3,19 @@ GRAPHIC_CARD_PKG=""
...
+        # Strip the path and leading X from the X11 socket
+        # but check that the resulting string is numeric and
+        # non-empty before exporting.
+        DISPLAY=:$(basename $i | sed -n -E 's/^X([0-9]+)/:\1/p')
```
nuclearsandwich, line 15: `$i` is no longer just the value of the `DISPLAY` variable so once confirming a likely active X11 server we need to get the value of `DISPLAY` for export. Compared with the prior substitution this one suppresses values that do not match `X[0-9]+`.

nuclearsandwich, line 15: I used `/` for the separator but if `@` is the preferred house style I can switch
```suggestion
        DISPLAY=:$(basename $i | sed -n -E 's@^X([0-9]+)@:\1@p')
```

jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -3,16 +3,19 @@ GRAPHIC_CARD_PKG=""
...
+        if [ -n $DISPLAY ]; then
+          export DISPLAY
+        fi
```
nuclearsandwich, line 18: Since the variable value could be empty, but an earlier socket file may have set a `DISPLAY`, only export the current value if it is not empty.

jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -80,6 +83,8 @@ if $GPU_SUPPORT_NEEDED; then
       echo "GPU support needed by the script but DISPLAY var is empty"
       # Try to restart lightdm. It should stop the script in the case of failure
       sudo service lightdm restart
+      # Wait for lightdm service to restart X11.
+      sleep 5
```
nuclearsandwich, line 87: 5 seconds is possibly way too long, but if this does not work as expected we likely need a different hypothesis.

jenkins-scripts/lib/check_graphic_card.bash
j-rivero: `/` is perfectly fine.

jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -80,6 +83,8 @@ if $GPU_SUPPORT_NEEDED; then
       echo "GPU support needed by the script but DISPLAY var is empty"
       # Try to restart lightdm. It should stop the script in the case of failure
       sudo service lightdm restart
```
j-rivero, line 85: Not sure if the systemd wrapper around `service` is still in place. We probably want to change it to use `systemctl`.

jenkins-scripts/lib/check_graphic_card.bash
nuclearsandwich: Yeah I'm not certain that the watching lightdm activation / status would be sufficient to resolve the race since I don't know if lightdm uses systemd's notification API for actually declaring readiness or if it's a simple/exec type service which is considered active as soon as the process runs https://www.freedesktop.org/software/systemd/man/systemd.service.html in which case we'd still need to wait ourselves for X11 to start.

jenkins-scripts/lib/check_graphic_card.bash
```diff
@@ -3,16 +3,19 @@ GRAPHIC_CARD_PKG=""
...
-      if [ $? -eq 0 ] ; then
-	export DISPLAY=$i
+      # If a process is running with the open socket then
+      # lsof will exit successfully.
+      if lsof -Fp $i ; then
+        # Strip the path and leading X from the X11 socket
+        # but check that the resulting string is numeric and
+        # non-empty before exporting.
+        DISPLAY=:$(basename $i | sed -n -E 's/^X([0-9]+)/:\1/p')
+        if [ -n $DISPLAY ]; then
```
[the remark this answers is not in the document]
nuclearsandwich: Good catch, I added quotes in 73ba462fc9db3fb183ec03e384300f49ee6159a8 I prefer POSIX compatible shell to Bash-isms where possible, although since this is an explicitly bash script I don't mind being overridden on this point.

jenkins-scripts/lib/check_graphic_card.bash
nuclearsandwich: I believe that it is, tested empirically, and I left this as is in order to retain support for past systems. But realistically our infra won't run on non-systemd-based hosts so there is no reason not to make the change which is in a2ff47eae866187254f91936e9a70b8005377029

PR #917, opened 2023-05-01
Create a Groovy script to apply singleton labels.

nuclearsandwich:
I think this script will need to be run as a system groovy script rather than with the sandbox. I have successfully tested it using the Jenkins script console.

I have not yet created the DSL job for it. I was hoping for some advice on how best to proceed there since storing the script in a standalone file isn't compatible with the system groovy script approach directly. I _think_ what I'll need to do is read the file out of the release-tools git checkout from the Groovy DSL and supply the contents as a string to the system groovy field on the job.

PR #918, opened 2023-05-08
Add the unstable parser and master label to the nightly_labeler

j-rivero:
Fixes a couple of problems with #917:

 * The `built-in` label is not ready in our combination of Jenkins version and node plugin version. Use `master` until we execute the full migration.
 * The build is not marked as unstable since it lacks the configuration to run the console-log parser that transform the `MARK_AS_UNSTABLE` text into real unstable build status.


j-rivero:
Need this deployed. Merging.

PR #925, opened 2023-05-17
Retry the build when `java.nio.channels.ClosedChannelException` found in failed log

Crola1702:
This PR adds a similar functionality as 9fbfe60 with the difference that is one is scoped to all the agents when `java.nio.channels.ClosedChannelException` is found in the build log.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFBase.groovy
@@ -38,5 +38,15 @@ class OSRFBase
           }
         }
       }
+      publishers {
+        configure { project ->
+          // When a build fails because an agent disconnects, retry it once
+          project / publishers / 'com.chikli.hudson.plugin.naginator.NaginatorPublisher' {
+            regexpForRerun("java.nio.channels.ClosedChannelException")
+            checkRegexp(true)
+            maxSchedule(1)
+          }
+        }
+      }
     }
 }


Crola1702:
I added an update feature so the child jobs don't override the previous regex:

https://github.com/Crola1702/release-tools/blob/d7634d83deebb56b3f08d401dea7fb81134adc2a/jenkins-scripts/dsl/_configs_/RetryFailures.groovy#L6

claraberendsen:
Great work @Crola1702!  I would probably add to this PR testing on the following scenarios: 
- Verify that when the agent disconnects the plugin still rebuilds the job. (idea for testing: take hold of an agent and add a shell build step that restarts the agent to break it). 
-  Verify that on a nvidia agent all the steps are still performed as expected (mainly the edge case here is if there is only one node available, that the rebuilding happens before the agent can be correctly applied the `recovery-process` label and the build fails again)

@j-rivero the justification on why to prefer this way instead of using another retry plugin follows the idea of maintaining one set of rules for retrial instead of having two separate actors that we need to take into account.  This has the benefit of ensuring that if the both errors happen in the same build the retrial of the job is only gonna get re triggered once, which is not a certainty when two different plugins are managing the rebuild, and would improve maintainability in the long term hopefully. 

The downside to this and an issue that it's not addressed in this PR, but needs to be addressed eventually, is that there is a shared logic on the other properties for rebuilding that may lead to side-effects. For example if we set on a child job(the unix for example) that the maxSchedule of the retries is 3, then this is going to be applied whenever a `Channel closed exception` happens. Even though this spill-off is not desirable, the only scenario we have at the moment is with the delay property, which would possibly add on nvidia agents an extra 70s delay to the rebuild when other errors happen.

I would argue that this is not a blocking behavior for this PR since it adds immediate value to the gazebo team and the side-effect is not breaking of the functionality.

Crola1702:
I was testing the first case Clara proposed. Here are the results:

In [_test_job_from_dsl](https://build.osrfoundation.org/job/_test_job_from_dsl/) I used my branch to check if when an agent disconnects, it reschedules the build

* For [_test_job_from_dsl/106/](https://build.osrfoundation.org/job/_test_job_from_dsl/106/) agent was disconnected. The build was rescheduled, and when the agent came back online I changed the label to `test-instance` and deleted the shutdown step.
* The result was [_test_job_from_dsl/107/](https://build.osrfoundation.org/job/_test_job_from_dsl/107/) that passed and has the `Re-scheduled after failure` mark

Crola1702:
@j-rivero. I think CI failed because I'm using a forked repository instead of an actual branch.

```
fatal: ambiguous argument 'origin/master...origin/retry-after-disconnection': unknown revision or path not in the working tree.
```

What should I do now? :thinking: Merge it into a different branch in gazebo-tooling repo?

j-rivero (changes requested):
Thanks @Crola1702 for the PR! And @claraberendsen for the comments.

 > @j-rivero the justification on why to prefer this way instead of using another retry plugin follows the idea of maintaining one set of rules for retrial instead of having two separate actors that we need to take into account. This has the benefit of ensuring that if the both errors happen in the same build the retrial of the job is only gonna get re triggered once, which is not a certainty when two different plugins are managing the rebuild, and would improve maintainability in the long term hopefully.

Sounds good to me. 

> * [update_retry_parameters](https://github.com/Crola1702/release-tools/blob/d7634d83deebb56b3f08d401dea7fb81134adc2a/jenkins-scripts/dsl/_configs_/RetryFailures.groovy#L6) will update the naginator parameters in the following way:   
>       * The new regex will be added before the previous regex. This means a new case is considered instead of just the child case.
>       * We can set a delay for the retry. It defaults to 0, and, in that case, it doesn't add a delay tag to xml configuration
>       * We can also set `checkRegexp` and `maxSchedule` parameters (they default to `true` and `maxSchedule` respectively

Some points on the design:

 * If we are going to use a new class to encapsulate some of the logic, I would prefer to move the configure code in `OSRFBase` into the new class, so we can use something like `RetryFailures.create(job, 'nvml error driver/library', 70)` directly.
 * If with this PR at some point we remove the code from the `OSRFBase` then we probably will face hidden bug not trivial to detect:
```xml
        <com.chikli.hudson.plugin.naginator.NaginatorPublisher>
            <regexpForRerun>|nvml error driver/library version mismatch</regexpForRerun>
            <checkRegexp>true</checkRegexp>
            <maxSchedule>1</maxSchedule>
            <delay class='com.chikli.hudson.plugin.naginator.FixedDelay'>
                <delay>70</delay>
            </delay>
        </com.chikli.hudson.plugin.naginator.NaginatorPublisher>
 ```
* I would prefer to use the same method as suggested above (`RetryFailures.create(job, 'nvml error driver/library', 70)`) for the updates on existing configurations and manage the difference in the arguments in some way:
  1. Make the code to fail if the values are not in sync. If you have defined a 70 secs for the delay and a new call used a different value, fail. That makes the consumers of the class to understand that they need a common value.
  1. Try to deal with the different values:  are we able to define a safe selection on `maxSchedule` and/or `delay` for  using as a default? That could be the case for `delay`, as @claraberendsen mentioned we can wait a bit longer without adding too much drama to the use cases I have in mind. `maxSchedule` is more complicated.

I'm fine with going with I. or II. handling delay to go to max + making `maxSchedule` to fail if values are different.
 
A couple of nitpicks:

 *  Can we rename the class to use a prefix like `Helper` or similar? All the classes in that dir are somehow prefixed.
 *  This is totally optional but for the class method arguments, maybe we can try to use the nice [named parameters](http://docs.groovy-lang.org/latest/html/documentation/#_named_parameters_2). Not that I have used them in the code that I remember of but they are so cool :)

**Comment by Crola1702** [comment]
> I would prefer to move the configure code in OSRFBase into the new class, so we can use something like RetryFailures.create(job, 'nvml error driver/library', 70) directly. 

:heavy_check_mark: 
---

>  would prefer to use the same method as suggested above (RetryFailures.create(job, 'nvml error driver/library', 70)) for the updates on existing configurations 

:heavy_check_mark: 
---

> Can we rename the class to use a prefix like Helper or similar? All the classes in that dir are somehow prefixed.

:heavy_check_mark: 
---

> If with this PR at some point we remove the code from the OSRFBase then we probably will face hidden bug not trivial to detect:

Thanks for noting that! The solution implemented on b1bd5a2 is:
* If no old value exists, just use the new one
* If an old value exists, append the new one using `|`

:heavy_check_mark: 
---

> This is totally optional but for the class method arguments, maybe we can try to use the nice [named parameters](http://docs.groovy-lang.org/latest/html/documentation/#_named_parameters_2). Not that I have used them in the code that I remember of but they are so cool :)

Thanks for the suggestion and the link. I tried to use them before but failed, now I'm using the named parameters. :heavy_check_mark: 

---

> manage the difference in the arguments in some way

I went for a combination of both options as @j-rivero  and @claraberendsen  suggested:
* For `delay` I'm using the highest of the 2 values
* For `checkRegexp` fails if values are different
* For `maxSchedule` fails if values are different

File: jenkins-scripts/dsl/_configs_/HelperRetryFailures.groovy [hunk 1]
```diff
@@ -0,0 +1,62 @@
+package _configs_
+
+import javaposse.jobdsl.dsl.Job
+
+class HelperRetryFailures {
+
+  /**@
+    * Create method for Nagiator Publisher retry plugin
```
j-rivero, line 8:
```suggestion
    * Create method for Naginator Publisher retry plugin
```

j-rivero, line 8: The documentation is great, thanks! Can we describe here what is going to happen with the different attributes if we try to call this method twice? That should help the user to know what exactly is going to happen.

jenkins-scripts/dsl/_configs_/HelperRetryFailures.groovy
```diff
@@ -0,0 +1,62 @@
...
+    * @param args.regexpForRerun - Regular expression to match against the console output
+    * @param args.checkRegexp - Whether to check the regular expression
+    * @param args.maxSchedule - Maximum number of times to retry the build
+    * @param args.delay - Delay in seconds to wait before retrying the build
+    */
+  static void create(Job job, Map args) {
+    job.with {
+      publishers {
+        configure { project ->
+          def old_regexpForRerun = (project / publishers / 'com.chikli.hudson.plugin.naginator.NaginatorPublisher' / regexpForRerun).text()
```
j-rivero, line 20: The code is a bit ugly with all the long selectors. Maybe we can use something like:
```groovy
          def naginator_root = project / publishers / 'com.chikli.hudson.plugin.naginator.NaginatorPublisher'
          
          def old_regexpForRerun = (naginator_root  / regexpForRerun).text()
 ```

And apply the root for the rest of the code.

**Review by j-rivero — CHANGES_REQUESTED** [review]
Wow, this has improved a lot. Let's do another round. I love the way of calling the function with the named parameters. Also great to use a standard documentation method for the class.

File: jenkins-scripts/dsl/_configs_/HelperRetryFailures.groovy (hunk 1 above)
Comment (Crola1702) [reply jenkins-scripts/dsl/_configs_/HelperRetryFailures.groovy:8]: I'd like to do that, but Naginator doesn't have documentation for that.

For example, yesterday I tried to look what will happen if `checkRegexp` is set to `false`. As I didn't find any documentation on the official site, I went to the source code and figured out that it's simply checking the regex that is defined above :upside_down_face: 

I'll add what I've figured out about each parameter.

**Comment by Crola1702** [comment]
After d3bb9d3. XML configurations were the same as before, so tests work the same as before :slightly_smiling_face:

File: jenkins-scripts/dsl/_configs_/HelperRetryFailures.groovy [hunk 3]
```diff
@@ -0,0 +1,66 @@
...
+          if (args.regexpForRerun) {
+            // If the old value exists, append the new value to it
+            def updated_regexpForRerun = old_regexpForRerun ? old_regexpForRerun + '|' + args.regexpForRerun : args.regexpForRerun
+            (naginator_root / regexpForRerun).value = updated_regexpForRerun
+          }
+
+          if (args.checkRegexp) {
+            // Fail if old and new checkRegexp are both set and different
+            if (old_checkRegexp && old_checkRegexp != args.checkRegexp) {
+              throw new Exception("checkRegexp is already set to ${old_checkRegexp} and cannot be changed to ${args.checkRegexp}")
```
j-rivero, line 44:
```suggestion
              throw new Exception("checkRegexp is already set to ${old_checkRegexp} and cannot be changed to ${args.checkRegexp}. Both values needs to be the same.")
```

jenkins-scripts/dsl/_configs_/HelperRetryFailures.groovy
```diff
@@ -0,0 +1,66 @@
...
+            }
+            (naginator_root / checkRegexp).value = args.checkRegexp
+          }
+
+          if (args.maxSchedule) {
+            // Fail if old and new maxSchedule are both set and different
+            if (old_maxSchedule && old_maxSchedule != args.maxSchedule) {
+              throw new Exception("maxSchedule is already set to ${old_maxSchedule} and cannot be changed to ${args.maxSchedule}")
```
j-rivero, line 52:
```suggestion
              throw new Exception("maxSchedule is already set to ${old_maxSchedule} and cannot be changed to ${args.maxSchedule}. Both values needs to be the same.")
```

Crola1702:
Thanks @j-rivero. This is ready to be merged. I have to ask you to do that because I don't have the permissions (Friendly ping @claraberendsen, maybe you can help me merge it)

Issue #929, opened 2023-05-22
Builds failing in drogon with error  `Did not run docker cleaner, level extreme not recognized`


claraberendsen:
### Description
Builds are failing with `unable to resolve host ip-172-30-1-203: Name or service not known` when we are trying to cleanup docker to free resources. 

### 
### Builds affected 
- [#1142](https://build.osrfoundation.org/job/gazebo-ci-pr_any-ubuntu_auto-amd64-gpu-nvidia/1142/console)
- [#1315](https://build.osrfoundation.org/job/gazebo-abichecker-any_to_any-ubuntu_auto-amd64/1315/console)
### Logs
<details>

```
echo 'Space left is low again: 87% used'
Space left is low again: 87% used
+++ echo 'Clean up the whole cache was not enough. Look for build/ directories bigger than :'
Clean up the whole cache was not enough. Look for build/ directories bigger than :
+++ source ./scripts/jenkins-scripts/docker/lib/docker_cleanup.sh extreme
++++ cleanup_level=extreme
+++++ sudo docker version --format '{{.Server.APIVersion}}'
sudo: unable to resolve host ip-172-30-1-203: Name or service not known
++++ docker_version=1.41
+++++ echo '1.41 > 1.25'
+++++ bc -l
++++ [[ -n 1 ]]
++++ case $1 in
++++ echo 'Did not run docker cleaner, level extreme not recognized (accepted values low | medium | high | extreme!! '
Did not run docker cleaner, level extreme not recognized (accepted values low | medium | high | extreme!! 
```

</details>

claraberendsen:
You might be right @j-rivero seems like the [legacy code](https://github.com/gazebo-tooling/release-tools/blob/f75c55a453848a9026fab1f4bd79fa05a2ec8fc5/jenkins-scripts/docker/lib/boilerplate_prepare.sh#LL46C1-L52C7) to extreme got into prod. Will remove that and try to do a test build with drogon. If that solves the issue I don't think we are going to need to bump the instances.

PR #930, opened 2023-05-22
remove legacy reference to 'extreme' option in docker cleanup process

claraberendsen:
### Description
This [issue](https://github.com/gazebo-tooling/release-tools/issues/929) appeared today due to legacy code expecting extreme option for the docker_cleanup process.

### DIFF
--- a/jenkins-scripts/docker/lib/boilerplate_prepare.sh
+++ b/jenkins-scripts/docker/lib/boilerplate_prepare.sh
@@ -43,13 +43,6 @@ if [[ -z ${DO_NOT_CHECK_DOCKER_DISK_USAGE} ]]; then
        source ${SCRIPT_DIR}/lib/docker_cleanup.sh high
     fi
 
-    # if not enough, try to clean up build/ directories
-    PERCENT_DISK_USED=$(df -h ${docker_device} | grep ${docker_device} | sed 's:.* \([0-9]*\)%.*:\1:')
-    if [[ $PERCENT_DISK_USED -gt 85 ]]; then
-        echo "Space left is low again: ${PERCENT_DISK_USED}% used"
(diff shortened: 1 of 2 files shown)


claraberendsen:
This doesn't solve the issue is that disk usage is over 90 % on drogon. It seems like there is limitation to the memory docker_cleanup script can do.  We are seeing this issue in drogon because it's not a machine that gets reprovisioned constantly. However this would unblock us from keeping drogon off the buildfarm.

jenkins-scripts/docker/lib/boilerplate_prepare.sh
```diff
@@ -43,13 +43,6 @@ if [[ -z ${DO_NOT_CHECK_DOCKER_DISK_USAGE} ]]; then
        source ${SCRIPT_DIR}/lib/docker_cleanup.sh high
     fi
```
j-rivero, line 45:
```suggestion
```

jenkins-scripts/docker/lib/docker_cleanup.sh
```diff
@@ -28,7 +28,7 @@ docker_version="$(sudo docker version --format '{{.Server.APIVersion}}')"
             sudo docker system prune --all -f
             ;;
         *)
-            echo "Did not run docker cleaner, level $cleanup_level not recognized (accepted values low | medium | high | extreme!! "
+            echo "Did not run docker cleaner, level $cleanup_level not recognized (accepted values low | medium | high !! "
```
j-rivero, line 31:
```suggestion
            echo "Did not run docker cleaner, level $cleanup_level not recognized (accepted values low | medium | high )"
```

j-rivero (approved):
Couple of cosmetic issues. Looks good.

PR #932, opened 2023-05-23
Change huge-memory to large-memory

Crola1702:
This PR changes node labels, so we're using the same convention for the agents that have a lot of memory to run heavy jobs.


j-rivero (approved):
Great, thanks!

PR #935, opened 2023-05-23
Set GITHUB_ACTIONS environment variable to fix homebrew

azeey:
This is a workaround per @scpeters' suggestion to fix homebrew bottle creation broken by https://github.com/Homebrew/brew/pull/15465


azeey:
@scpeters , is there a way I can test this without merging?

j-rivero:
> @scpeters , is there a way I can test this without merging?

Probably not needed anymore, see https://github.com/Homebrew/brew/pull/15465#issuecomment-1560975707. Closing.

scpeters:
agreed; I think this isn't needed anymore

PR #948, opened 2023-06-02
Make nightly_labeler to run during the nightly generation

### LINKED ISSUE #946: `linux-nightly-focal` labels don't have agents to run
There are a lot of debbuilders that wait for a node with `linux-nightly-focal` label to become available. Today, there wasn't an agent with this label in the buildfarm, so I manually added it to linux-ec2 instance.

List of jobs that were waiting:
* https://build.osrfoundation.org/job/gz-tools2-debbuilder/
* https://build.osrfoundation.org/job/gz-utils2-debbuilder/
* https://build.osrfoundation.org/job/gz-math7-debbuilder/
* https://build.osrfoundation.org/job/gz-plugin2-debbuilder/
* https://build.osrfoundation.org/job/gz-common5-debbuilder/
* https://build.osrfoundation.org/job/gz-msgs10-debbuilder/
* https://build.osrfoundation.org/job/gz-rendering8-debbuilder/
* https://build.osrfoundation.org/job/sdformat13-debbuilder/
* https://build.osrfoundation.org/job/gz-fuel-tools9-debbuilder/
* https://build.osrfoundation.org/job/gz-transport13-debbuilder/
* https://build.osrfoundation.org/job/gz-gui8-debbuilder/
* https://build.osrfoundation.org/job/gz-sensors8-debbuilder/
* https://build.osrfoundation.org/job/gz-sim8-debbuilder/

j-rivero:
Should fix #946

### DIFF
--- a/jenkins-scripts/dsl/_configs_/Globals.groovy
+++ b/jenkins-scripts/dsl/_configs_/Globals.groovy
@@ -12,8 +12,11 @@ class Globals
 
    static CRON_EVERY_THREE_DAYS = 'H H * * H/3'
    static CRON_ON_WEEKEND = 'H H * * 6-7'
-   static CRON_PREPARE_NIGHTLY = '45 8 * * *'
    static CRON_START_NIGHTLY = '0 9 * * *'
+   // Run nightly scheduler during the nightly creation to be sure
+   // that any possible node killed is replaced. Starting -15min
+   // before CRON_NIGHTLY_NODES and evert 20min for 3 hours
+   static CRON_NIGHTLY_NODES = '45/20 8-11 * * *'
 
    // Only one -E regex can be passed, so make a regex that matches both
    // _ign_TEST and _gz_TEST
--- a/jenkins-scripts/dsl/core.dsl
+++ b/jenkins-scripts/dsl/core.dsl
@@ -105,7 +105,7 @@ nightly_labeler.with
   label Globals.nontest_label("master")
 
   triggers {
-    cron(Globals.CRON_PREPARE_NIGHTLY)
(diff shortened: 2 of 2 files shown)


Crola1702:
Wouldn't be easier to add these labels to the agent configuration? Then we will have a nightly label always available

j-rivero:
> Wouldn't be easier to add these labels to the agent configuration? Then we will have a nightly label always available

That would be easier indeed, we had that when we had the in-office agents. The problem now is that we rely on agents on AWS that are sometimes destroyed and new ones are created automatically from a template so having individual labels inside the agents created with templates is not easy. That is the reason why we have this dynamic setup approach.

Crola1702:
Okay, makes sense. I was thinking in the chef approach. Chaning labels in [osrf_jenkins_agent default recipe](https://github.com/osrf/osrf_jenkins_agent/blob/5005ba6bad0a93f233916d42151883cfa830cb43/recipes/default.rb#L54-L56)

Is there any reason we don't want this label to be in all the nodes?

j-rivero:
> I was thinking in the chef approach. Chaning labels in [osrf_jenkins_agent default recipe](https://github.com/osrf/osrf_jenkins_agent/blob/5005ba6bad0a93f233916d42151883cfa830cb43/recipes/default.rb#L54-L56)
> 
> Is there any reason we don't want this label to be in all the nodes?

To be able to build all the packages inside a given Gazebo collection in a topological order we need to orchestrate their builds somehow. The approach we are using with the nightlies is to restrict the node builders to just one so we can launch the nightlies in order and they will be process in that order. If we use chef and have more than one node with the same label we'll get parallel builds and the order will not be maintained.

Issue #895, opened 2023-03-29
Support for releasing ros_gzgarden unofficial packages for Humble


j-rivero:
## Motivation

Gz Garden [is not currently supported by binaries in any ROS 2 distribution](https://gazebosim.org/docs/garden/ros_installation#gazebo-garden-not-recommended-) although the `ros_gz` bridge supports from source compilations using Gz Garden for different ROS 2 distributions. Building from source is the recommended way for advanced users that really want to use Garden on ROS 2 Humble or ROS 2 Rolling.

For Gazebo Classic with its 11 version and ROS 1 in this situation where the official version was an older version of Gazebo,  we generated years ago **unofficial** packages named `gazebo11_ros_pkgs` hosted in `packages.osrfoundation.org` that could be used to [install Gazebo11 instead of Gazebo9 on Melodic and others](https://classic.gazebosim.org/tutorials?tut=ros_wrapper_versions&cat=connect_ros#Gazebo11.xseries). 

In this case, we can try to generate **unofficial** packages for the users of Gazebo 
Garden + ROS 2 Humble that could help the advanced user with their installations. 

## Roadmap:

 * [x] Support for rosdep to rename Gz Fortress to Gz Garden
    * https://github.com/osrf/osrf-rosdep/pull/12 
    * https://github.com/osrf/osrf-rosdep/pull/13
 * [x] Jenkins DSL support for debbuilders of ros_gz
    * https://github.com/gazebo-tooling/release-tools/pull/896 
    * https://github.com/gazebo-tooling/release-tools/pull/897
 * [x] Scripts to support forking ros-gbp2 release, setup the fork and releasing operations
    * https://github.com/gazebo-tooling/release-tools/pull/898
    * https://github.com/gazebo-tooling/release-tools/pull/926
    * https://github.com/gazebo-tooling/release-tools/pull/927
 * [x] Release a test version into the prerelease repository
 * [x] Create the final fork repository in GitHub osrf
   * https://github.com/gazebo-release/ros_ign-release 
 * [x] Release into stable from the fork repository
   * Tested https://build.osrfoundation.org/job/ros_gz_bridge-install-pkg_unofficial-any-manual/43/ 
 * [x] Periodic smoke test to detect possible ABI breakages if ROS official packages make any ABI incompatible change 
   *  https://github.com/gazebo-tooling/release-tools/pull/950 
   *  https://github.com/gazebo-tooling/release-tools/pull/956

## Testing:

[...]

PR #950, opened 2023-06-02
Unofficial ros_gz: add a smoke test to detect ABI breakages

j-rivero:
Part of #895

The pull request add an smoke test to the buildfarm that is designed to mainly identify extreme breakages in the software with the eyes set on any possible ABI disruptive change that ROS dependencies of the wrappers could execute. To do so, the testing does:

 * Reproduce ros_gz talker/listener and check for expected output
 * Execute `ros2 launch ros_gz_sim gz_sim.launch.py gz_args:=shapes.sdf` and leave it for 3 min checking that indeed it has been running for 3min.

Not very serious tests but should serve us to detect heavy problems.

Bonus: remove of an old file in this PR not used anymore.


j-rivero:
Tested: [Build Status](https://build.osrfoundation.org/job/ros_gz_bridge-install-pkg_unofficial-any-manual/41/)

PR #959, opened 2023-06-16
Fix nvml error discovery in log

Crola1702:
`nvml error: driver/library version mismatch` was missing a `:` character in the `regexpForRerun` attribute. This PR adds that character

I also moved the "# BEGIN SECTION: NVIDIA MISMATCH RECOVERY" to the real section where the recovery is made, replacing the original message with "Checking if nvidia mismatch error is present in log" for debug processes

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
@@ -53,11 +53,12 @@ class OSRFUNIXBase extends OSRFBase
                   def node = build.getBuiltOn()
                   def old_labels = node.getLabelString()
 
-                  println("# BEGIN SECTION: NVIDIA MISMATCH RECOVERY")
+                  println("Checking if nvidia mismatch error is present in log")
                   if (!(build.getLog(1000) =~ "nvml error: driver/library version mismatch")) {
                     println(" NVIDIA driver/library version mismatch not detected in the log - Not performing any recovery automatic recovery step")
                     return 1;
                   } else {
+                    println("# BEGIN SECTION: NVIDIA MISMATCH RECOVERY")
                     try {
                       println(" PROBLEM: NVIDIA driver/library version mismatch was detected in the log. Try to automatically resolve it:")
                       println("Removing labels and adding 'recovery-process' label to node")
@@ -85,7 +86,7 @@ class OSRFUNIXBase extends OSRFBase
     // Add the new regex to naginator tag
     // There is no need to specify checkRegexp and maxSchedule because they are the default values
     HelperRetryFailures.create(job, [
-      regexpForRerun: "nvml error driver/library version mismatch",
+      regexpForRerun: "nvml error: driver/library version mismatch",
       delay: 70
     ])
   }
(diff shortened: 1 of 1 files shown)


jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -53,11 +53,12 @@ class OSRFUNIXBase extends OSRFBase
                   def node = build.getBuiltOn()
                   def old_labels = node.getLabelString()
 
-                  println("# BEGIN SECTION: NVIDIA MISMATCH RECOVERY")
```
claraberendsen, line 56: (nit) I think is worthwhile to add a section to say that it's a postbuild recovery process. Maybe something in the lines of `#BEGIN SECTION: ATTEMPTING TO RECOVER FROM FAILURE`

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
```diff
@@ -53,11 +53,12 @@ class OSRFUNIXBase extends OSRFBase
...
+                  println("Checking if nvidia mismatch error is present in log")
                   if (!(build.getLog(1000) =~ "nvml error: driver/library version mismatch")) {
                     println(" NVIDIA driver/library version mismatch not detected in the log - Not performing any recovery automatic recovery step")
```
claraberendsen, line 58: Also an while we are at it, fixing typos
```suggestion
                    println(" NVIDIA driver/library version mismatch not detected in the log - Not performing any automatic recovery step")
```

claraberendsen (approved):

claraberendsen:
Can you add here the build that proves that this solves the issue? @Crola1702

Crola1702:
> Can you add here the build that proves that this solves the issue? @Crola1702

<details>

<summary>

### Jenkins Agent disconnection

</summary>

The configuration for these builds was adding `echo "java.nio.channels.ClosedChannelException"` and `exit 1`, so the build shows the error and then fails.

Build [130](https://build.osrfoundation.org/job/_test_job_from_dsl/130/) was triggered with this configuration. As a result, build [131](https://build.osrfoundation.org/job/_test_job_from_dsl/131/) was retriggered by naginator.

</details>

<details>

<summary>

### NVML error

</summary>

The configuration for this builds was adding:
```
cat << EOF
nvml error: driver/library version mismatch
EOF

exit 1 
```

so the build shows the error and then fails.

Build [132](https://build.osrfoundation.org/job/_test_job_from_dsl/132/) was triggered with this configuration. As a result, build [133](https://build.osrfoundation.org/job/_test_job_from_dsl/133/) was retriggered by naginator and the machine was restarted (as I don't have evidence of this, I ran again the configuration with [137](https://build.osrfoundation.org/job/_test_job_from_dsl/137/) and its rebuild: [138](https://build.osrfoundation.org/job/_test_job_from_dsl/138/) )

image

image
> Because the instance labels were reloaded

</details>

jenkins-scripts/dsl/_configs_/OSRFUNIXBase.groovy
Crola1702: I moved down the `#BEGIN SECTION` as it would make sense that the section is "recovering process" instead of "checking if I can do a recovery process".

PR #965, opened 2023-06-30
Fix outdated job cron trigger

Crola1702:
Currently, [_outdated_job_runner](https://build.osrfoundation.org/job/_outdated_job_runner/) it's running every minute instead of every hour as expected.


Crola1702:
I'll merge this, as the current Jenkins configuration already has this change (and the job has run 15 times)

nuclearsandwich (commented):
✅ Definitely cron strings are a thing to review closely! This is a needed fix.

PR #962, opened 2023-06-21
Add new job to run outdated gazebo jenkins jobs

Crola1702:
### Description

This PR adds a new job that will automate the manual process of running outdated jobs.

The automation this new job adds will clear all human error to maintain the gazebo buildfarm jobs up to date by running every night and checking all the jobs buildfarmers are tracking.

After merging this PR, there will be a new job called `_outdated_job_runner`. It will run nightly (as defined in [`CRON_NIGHTLY_NODES`](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/_configs_/Globals.groovy#L18)), and it will trigger a maximun of 7 jobs per operating system to prevent buildfarm overload if more jobs are outdated (this was internally defined by the buildfarmer team).

<details>

<summary>
Some extra context:
</summary>

Currently, Gazebo jobs only run with `scm` triggers. Sometimes, the repositories don't change in a lot of time, and we don't know if they're still working as they should (the main concern are dependencies. A broken python package can lead to many problems if not noticed early). Right now, we have an 8 days threshold for a job to be considered "outdated". When this happens, buildfarmer daily report (e.g., [2023-06-21.md](https://github.com/osrf/buildfarmer/blob/main/2023/25/2023-06-21.md)) will show a message saying `JOB HAS NOT BEEN UPDATED RECENTLY`. Then, we check and rebuild all outdated jobs by hand.

</details>

### Testing

The generated job was named [`_nightly_job_runner`](https://build.osrfoundation.org/job/_nightly_job_runner/) before. It had some test configurations that I used to verify that it was running correctly. For example: 
* Build to check what were the outdated jobs ([_nightly_job_runner#17](https://build.osrfoundation.org/job/_nightly_job_runner/17/console))
* Triggered build by _nightly_job_runner: [ign_gui-ign-6-win#75](https://build.osrfoundation.org/job/ign_gui-ign-6-win/75/) (you can see the upstream project that built it and all the rebuilds before it

Now I'm using [`_outdated_job_runner`](https://build.osrfoundation.org/job/_outdated_job_runner/) to test the new configuration:
* Build that triggered a `_test_job_from_dsl` build ([_outdated_job_runner#4](https://build.osrfoundation.org/job/_outdated_job_runner/4/))
  <details>
  
  I used the following script to trigger a `_test_job_from_dsl` build:

  ```
  def job = "_test_job_from_dsl";

  println('Running: ' + job)
  runJob(job)
  ```

  </details>
[...]

PR #966, opened 2023-07-03
Fix `CRON_HOURLY` string

Crola1702:
CRON_HOURLY was unintentionally defined to run every minute in #962 

After #965 it was, again, unintentionally defined to run every day instead of every hour.

This PR corrects this problem by making sure it runs every hour using the correct syntax.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/Globals.groovy
+++ b/jenkins-scripts/dsl/_configs_/Globals.groovy
@@ -11,7 +11,7 @@ class Globals
    static gazebodistro_branch = false
 
    static CRON_EVERY_THREE_DAYS = 'H H * * H/3'
-   static CRON_HOURLY = 'H H * * *'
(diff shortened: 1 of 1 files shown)


jenkins-scripts/dsl/_configs_/Globals.groovy
```diff
@@ -11,7 +11,7 @@ class Globals
    static gazebodistro_branch = false
 
    static CRON_EVERY_THREE_DAYS = 'H H * * H/3'
-   static CRON_HOURLY = 'H H * * *'
+   static CRON_HOURLY = 'H * * * *'
```
nuclearsandwich, line 14: This string will run at an arbitrary minute of the hour `H`, of any hour of the day, any day of the month, any month of the year, any day of the week.

nuclearsandwich (approved):
Third time lucky.

Issue #967, opened 2023-07-07
Memory problems with nightly builds using linux-nightly-* label on simple nodes


j-rivero:
One of the nightly builds of gz-physics failed probably because the node that built it did not have enough memory. That is usually fixed by the use of `large-memory` label in the gz-physics debbuilders but for nightlies we are implicitly injecting the `linux-nightly-*` node label so that configuration is not being respected.

A workaround would be to use `large-memory` nodes as the nightly builders so we are sure that we don't suffer from memory problems. The long-term solution is the proper orchestration of the build that respect the node labels in the debbuilders and can trigger the different package build in order.

PR #969, opened 2023-07-12
Use large-memory nodes for nightlies

j-rivero:
Closes: #967 

Modified the code (and comments) to use the label `large-memory` instead of `docker`. I also found not true that current code is able to support label expression, looking at the code `node.getLabelString().contains()` is probably unable to parser that expressions. Anyway, the filter of test instances is being done inside `findAll`.

I also removed outdate info in 7767c4f8e7715a999b1697beeecf4ef1ce0c0ae5

### DIFF
--- a/jenkins-scripts/tools/label-assignment-backstop.groovy
+++ b/jenkins-scripts/tools/label-assignment-backstop.groovy
@@ -11,16 +11,13 @@ import jenkins.model.*;
 import hudson.model.Label;
 
 /*
-  Each label is in a 2-tuple with a label expression that indicates its "pool"
-  of potential assignees.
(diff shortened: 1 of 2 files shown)


jenkins-scripts/tools/label-assignment-backstop.groovy
```diff
@@ -52,7 +50,6 @@ for (tup in exactly_one_labels) {
       node.computer.online &&
       node.getLabelString().contains(pool_label) &&
       !node.getLabelString().contains(nightly_label_prefix) &&
-      !node.getLabelString().contains("gpu-nvidia") &&
```
Crola1702, line 55: Does this mean all agents will be treated equally?

PR #970, opened 2023-07-14
Change gz-collections.yaml spec to support ci configs 

j-rivero:
The PR should not change any existing code but reforms the yaml spec to define how the libs and ci interacts:
 * 872d0df8a2080f707241cc900b7f0d19e54977b7 modifies the yaml to include a `ci_configs` section that can host all related CI details and the different collections only reference them by name. Names are equal to Ubuntu distributions but could be something different.
 * eb54989f12460608c14743ff1831876d8b64872d modifies gazebo_libs by doing some changes:
   *  `generate_ciconfigs_by_lib` is a parse step that builds an index of ci-configurations supported by each lib. It should help to avoid extra parsing in the diffferent jobs creation.
   * `generate_label_by_requirements` handles the label selection for the jobs, nvidia or large-memory.
   * The code is not handling the exclude section in ci
   * Rest of the changes are just to adapt the job creation iteration to the new index.

I would not expect any XML code change with the exception of gz-harmonic repo removal since now we are excluding it.


j-rivero:
Oh ignition_collection needs to be migrated, back to draft.

j-rivero:
> Oh ignition_collection needs to be migrated, back to draft.

ok, ready. CI says that there is no contect change in any job or any new job.

j-rivero:
There is no change in the final XML configurations according to CI, I'm going to merge the PR to create new ones.

PR #972, opened 2023-07-25
Update outdated job condition from 8 to 4 days

Crola1702:
The 8 days threshold from an outdated job was defined when we ran the gazebo jobs manually using daily report. Now that this work is automated, it would make sense to decrease the threshold, so the gazebo jobs are run more frequently by [_outdated_job_runner](https://build.osrfoundation.org/job/_outdated_job_runner/).

CC: @Blast545 @claraberendsen

### DIFF
--- a/jenkins-scripts/tools/outdated-job-runner.groovy
+++ b/jenkins-scripts/tools/outdated-job-runner.groovy
@@ -226,7 +226,7 @@ def jenkinsJobs = Hudson.instance
 
 def jobsToRun = [osx: [], win: [], docker: []]
 
-long eightDaysAgoMillis = System.currentTimeMillis() - 8 * 24 * 60 * 60 * 1000; // 8 days ago in milis
+long eightDaysAgoMillis = System.currentTimeMillis() - 4 * 24 * 60 * 60 * 1000; // 4 days ago in milis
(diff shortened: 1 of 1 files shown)


j-rivero:
I've have nothing against changing it but let's keep in mind the consequences: a bit more work load for the buildfarm (I think that it can support it given that we did not see any long queues in a while) and a bit more of false positives due to infra issues or downtime in services used by infra. In exchange we gain shorter testing cycles when dependencies or infra changes.

Crola1702:
>  bit more of false positives due to infra issues or downtime in services used by infra

I'm not sure what you mean by false positives, and infra issues failing ratio should remain the same. Besides that, I think that any new or known issue we could get from shorter testing cycles will improve the quality of our investigations as we could have more data to work with.

j-rivero:
>  I'm not sure what you mean by false positives, 

A false positive is a build that is failing but the reason is not a problem in the code that needs to be fixed.

> and infra issues failing ratio should remain the same.

If run the CI builds more often I would expect to find more problems related to infra issues like 404 to external services, problems with packages repositories being synced or other things not related to code testing.

> I think that any new or known issue we could get from shorter testing cycles will improve the quality of our investigations as we could have more data to work with.

I agree with this. Let's try merging and see if our mechanism of injecting builds keeps the server out of waiting queues.

PR #983, opened 2023-08-03
Fix sdformat14 dsl

azeey:
#982 added `sdformat14` in `ignition.dsl`, which I think should be `sdformat.dsl` instead.


azeey:
diff:
```
Only in /tmp/current_xml_configuration: gz-sdformat14-debbuilder.xml
Only in /tmp/pr_xml_configuration: sdformat14-debbuilder.xml
```

j-rivero (approved):

PR #987, opened 2023-08-08
Remove duplicate `large-memory` label

Crola1702:
@Blast545 noticed that this document had duplicated the `large-memory` label. I think the one that remains is the most clear and concise between the two definitions.


Blast545 (approved):
Maybe you can combine the two definitions? but that's a nitpick, lgtm. Let's wait for codeowners to comment before merging.

j-rivero:
> Maybe you can combine the two definitions? but that's a nitpick, lgtm. Let's wait for codeowners to comment before merging.

I think that we good with the changes.

PR #991, opened 2023-08-23
Change caching dir to default one 

claraberendsen:
### Description
This PR modifies the location of the cache files in the agent to mirror the default configuration where vcpkg builds it. The goal of this is to use the baked AMI cache instead of rebuilding it every time a new agent comes to life. 

#### Build test with new baked AMI
[Build Status](https://build.osrfoundation.org/job/test_ignition_common-pr-win/7/)


nuclearsandwich (commented):
Retroactive +1

PR #999, opened 2023-08-30
Generalize ros_gz scripts to support other releases than Garden

j-rivero:
Change the scripts to handle the upcoming Harmonic release by being able to deal with more than Garden releases.

Mostly change Garden hardcoded by a parameter and pass it all around the different bash + docker machinery but also stop hardcoding the osrf/rosdep URL for replacing values by a parameter.

Finally update the README to explain how to invoke the scripts with the current example of Garden.


azeey (approved):
Did a quick pass and looks good.

PR #1004, opened 2023-09-01
Install gz-sim*-cli on ros_gz testing builds

j-rivero:
The `ros_gz_sim` package for being able to run with ros launch needs to have the `gz-sim*-cli` package. This is probably a bug in `ros_gz` code.

Until we get that fixed, this PR adds automatically that package into the testing job so the smoke test can be run.

Tested: [Build Status](https://build.osrfoundation.org/job/ros_gz_bridge-install-pkg_unofficial-any-manual/55/)

### DIFF
--- a/jenkins-scripts/docker/ros_gz-install-test-job.bash
+++ b/jenkins-scripts/docker/ros_gz-install-test-job.bash
@@ -30,6 +30,10 @@ fi
 echo '# END SECTION'
 
 echo '# BEGIN SECTION: test gz_sim via ros2 launch'
+# Workaround until ros_gz defines a dependency on gz-sim*-cli. Avoid to use
+# apt --install-suggests since it installs the whole universe of packages
+VER=dpkg -l | grep libgz-sim | grep ^ii | head -1 | awk '{ print \$3 '}
+apt-get install -y gz-sim\${VER:0:1}-cli
 TEST_START=\`date +%s\`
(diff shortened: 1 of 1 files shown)


jenkins-scripts/docker/ros_gz-install-test-job.bash
```diff
@@ -30,6 +30,10 @@ fi
 echo '# END SECTION'
 
 echo '# BEGIN SECTION: test gz_sim via ros2 launch'
+# Workaround until ros_gz defines a dependency on gz-sim*-cli. Avoid to use
+# apt --install-suggests since it installs the whole universe of packages
+VER=\$(dpkg -l | grep libgz-sim | grep ^ii | head -1 | awk '{ print \$3 '})
```
azeey, line 35: Not sure why there's a `\$` instead of `$` and in the `awk '{ print \$3 '}`, should the `'` be after `}`?

j-rivero: > Not sure why there's a \$ instead of $
An extra `\` is needed since we are in a bash `"""` block that is going to be consumed by bash, if we don't use that this will be translated to a bash variable instead of reaching the real execution as pure `$3`. 

> awk '{ print \$3 '}, should the ' be after }?

Umm this indeed seems a bug that bash passed just fine. I change it.

PR #1007, opened 2023-09-07
homebrew_formula_pullrequest.bash: sanitize ~

scpeters:
Sanitize ~ characters from the brew version string.

Follow up to https://github.com/osrf/homebrew-simulation/issues/2390.

### DIFF
--- a/jenkins-scripts/lib/homebrew_formula_pullrequest.bash
+++ b/jenkins-scripts/lib/homebrew_formula_pullrequest.bash
@@ -26,6 +26,10 @@ if [ -z "${SOURCE_TARBALL_SHA}" ]; then
 fi
 echo '# END SECTION'
 
+# brew does not support ~ in VERSION string, so convert it to -
+VERSION_SANITIZED=$(echo "${VERSION}" | tr '~' '-')
+echo Sanitizing version string ${VERSION} to ${VERSION_SANITIZED}
+
 PULL_REQUEST_HEAD_REPO=git@github.com:osrfbuild/homebrew-simulation.git
 
 . ${SCRIPT_LIBDIR}/_homebrew_github_setup.bash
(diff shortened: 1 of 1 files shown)


scpeters:
testing with [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1400/) https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1400/

scpeters:
> testing with [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1400/) https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1400/
> 
> which is a manual re-run of https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1391 with the `VERSION` field changed from `1.0.0~pre2` to `1.0.0~pre3`

the resulting pull request ( https://github.com/osrf/homebrew-simulation/pull/2393 ) uses a version string with `-` instead of `~` as intended, and brew CI appears happy, so I consider this a successful test

j-rivero:
The point for using `~` in the Debian versions is to get `x.y.z~` resolved as lower than `x.y.z` so we can use `x.y.z~pre1` before `x.y.z`. I can not find any reference about what system is brew using for comparing versions but we would need that behavior to be preserved when changing the `~`

scpeters:
> The point for using `~` in the Debian versions is to get `x.y.z~` resolved as lower than `x.y.z` so we can use `x.y.z~pre1` before `x.y.z`. I can not find any reference about what system is brew using for comparing versions but we would need that behavior to be preserved when changing the `~`

`brew` uses custom logic in its [version.rb](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/version.rb) file. I've tested it a bit interactively using `brew ruby -e`, and it looks like it works. I'll post some output from these commands in a separate comment

scpeters:
~~~
$ brew ruby -e 'puts Version.new("1.0") == Version.new("1.0")'
true
$ brew ruby -e 'puts Version.new("1.0") == Version.new("1.0.0")'
true
$ brew ruby -e 'puts Version.new("1.0") == Version.new("1.0.0.1")'
false
$ brew ruby -e 'puts Version.new("1.0") == Version.new("1.0-pre1")'
false
$ brew ruby -e 'puts Version.new("1.0") > Version.new("1.0-pre1")'
true
~~~

scpeters:
> The point for using `~` in the Debian versions is to get `x.y.z~` resolved as lower than `x.y.z` so we can use `x.y.z~pre1` before `x.y.z`. I can not find any reference about what system is brew using for comparing versions but we would need that behavior to be preserved when changing the `~`

@j-rivero I believe I've responded to this comment

j-rivero:
> > The point for using `~` in the Debian versions is to get `x.y.z~` resolved as lower than `x.y.z` so we can use `x.y.z~pre1` before `x.y.z`. I can not find any reference about what system is brew using for comparing versions but we would need that behavior to be preserved when changing the `~`
> 
> @j-rivero I believe I've responded to this comment

Thanks for checking,  seems fine to me.

PR #1012, opened 2023-09-14
Make previous version optional in source_changelog.bash

azeey:
The latest tagged version is used as the default.

Not sure if this is useful for everyone, but it's been handy when generating changelogs for multiple repos.


scpeters:
it looks like CI fails for pull requests from a fork

scpeters (approved):

scpeters:
> it looks like CI fails for pull requests from a fork

see https://github.com/gazebo-tooling/release-tools/pull/1021

j-rivero (approved):

PR #1013, opened 2023-09-14
Make previous version optional in release_pull_request.bash

azeey:
This also adds some documentation

### DIFF
--- a/source-repo-scripts/release_pull_request.bash
+++ b/source-repo-scripts/release_pull_request.bash
@@ -14,27 +14,32 @@
 # limitations under the License.
 #
 
-# The script will open a pull request for a forward port.
+# The script will open a pull request a release PR (actually opens the PR page in your web browser)
 #
 # Requires the 'gh' CLI to be installed.
 #
 # Usage:
-# $ ./merge_forward_pull_request.bash <from_branch> <to_branch>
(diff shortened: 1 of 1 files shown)


source-repo-scripts/release_pull_request.bash
```diff
@@ -14,28 +14,33 @@
...
-  echo "./release_pull_request.bash <version> <to_branch> <previous_version>"
+if [[ $# -ne 2 ]]; then
+  echo "./release_pull_request.bash <version> <to_branch>"
   exit 1
 fi
 
 set -e
 
+git fetch --tags
+PREV_VER=$(git describe --tags --abbrev=0 | sed 's/.*_//')
```
scpeters, line 42: I'm tempted to try to use the `TO_BRANCH` variable here to try to make sure we identify a tag that matches the major version of our target branch, though it may not be necessary. Feel free to ignore this suggestion. 

```suggestion
PREV_VER=$(git describe --tags --match ${TO_BRANCH}* --abbrev=0 | sed 's/.*_//')
```

scpeters (approved):

source-repo-scripts/release_pull_request.bash
j-rivero: +1

PR #1014, opened 2023-09-15
Big clean up of release.py script

j-rivero:
A big first clean up in the ancient `release.py` script. Some major changes that I believe probably noone was using nowadays:

- Remove `--no-ignition-auto` option
- Remove `--upstream` option
- Remove i386 from release.py (Bionic is the last Ubuntu and we are probably not interested into Debian/i386)
- For create_tarball_name remove the option of package_alias (broken)
- Remove sanity_check_gazebo_versions (outdated)

Formatting/Linting fixes:

- Remove unused variables
- 2 lines after methods
- Fix spacing and blank lines
- Use byte stream for empty return type in check_call
- Change method not to end up in exit

The changes are easier to navigate commit-by-commit, the commit description should be good to understand each of them.

### DIFF
--- a/release.py
+++ b/release.py
@@ -281,19 +281,6 @@ def sanity_package_version(repo_dir, version, release_version):
     print_success("Package release versions in changelog")
 
 
-def sanity_check_gazebo_versions(package, version):
-    if package == 'gazebo':
-        if int(version[0]) > 1:
-            error("Error in gazebo version. Please use 'gazebo-current' package for gazebo 2")
-    elif package == 'gazebo-current':
-        if int(version[0]) < 2:
-            error("Error in gazebo-current version. Please use 'gazebo' package for gazebo 1.x")
-    else:
-        return
-
-    print_success("Gazebo version in proper gazebo package")
-
-
 def sanity_check_sdformat_versions(package, version):
     if package == 'sdformat':
         if int(version[0]) > 1:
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -493,7 +495,7 @@ def generate_upload_tarball(args):
     # If we're releasing under a different name, then rename the tarball (the
     # package itself doesn't know anything about this).
     if args.package != args.package_alias:
-        tarball_fname = '%s-%s.tar.bz2'%(args.package_alias, args.version)
+        tarball_fname = ' %s-%s.tar.bz2' % (args.package_alias, args.version)
```
azeey, line 498: This failed for me with
```

 !! Can not find a tarball at: /tmp/tmp8ttiifnf/build/ gz-cmake-3.4.1.tar.bz2 or at /tmp/tmp8ttiifnf/build/ gz_cmake-3.4.1.tar.bz2

```

I think the issue is there are a number of places where an extra space is added inside string formats that didn't have it before. Search for `' %s`

azeey (changes requested):

release.py
j-rivero: Good catch. Fixed in [38f04ab](https://github.com/gazebo-tooling/release-tools/pull/1014/commits/38f04ab0dcd479b13edd1906c311cc7418a86909) and [f82e937](https://github.com/gazebo-tooling/release-tools/pull/1014/commits/f82e937ec0b9920f81fcaeb691b764dc271d3b86)

PR #1017, opened 2023-09-22
Include all gazebo releases in gz-collection.yaml

j-rivero:
With the goal of implementing the `-ci-pr_any_` and `-abichecker-pr_any-` (needs to be done automically for all the Gazebo releases since these jobs handle all supported branches under the same name) this PR adds support for all the Gazebo releases supported, adding Citadel, Fortress and Garden to the existing Harmonic and including a new `__upcoming__` to represent the new release in development and support CI on the new branches or new platforms (if any).

The code still has the limitation of only generating jobs for Harmonic with the goal of controlling the changes and deliver them gradually in the upcoming PRs. In this PR, the DSL generation should not change anything but generating a dashboard for `__upcoming__`.


j-rivero:
The CI reports:
```
Only in /tmp/pr_xml_configuration: __upcoming__.xml
```

This is going to generate only a new dashboard called `__upcoming__`

Crola1702:
@j-rivero. I'm trying to fix `common.py`  again. It seems that is failing to retrive `gz-fuel-tools` from `__upcoming__` release. Is this intended? Or are we missing different repositories?

j-rivero:
> @j-rivero. I'm trying to fix `common.py` again. It seems that is failing to retrieve `gz-fuel-tools` from `__upcoming__` release. Is this intended? Or are we missing different repositories?

Is there any `gz-fuel-tools` listed under `__upcoming__`? Should be a valid use case not to have all the libraries listed in a release, specially in the `__upcoming__` one.

Crola1702:
Is the `__upcoming__` release pointing just to the new package releases and not showing old ones?

j-rivero:
> Is the `__upcoming__` release pointing just to the new package releases and not showing old ones?

The releases entries can be form by 1 to N repositories of code not having to be the same than the old ones. There is no guarantee that one of the new releases include all the previous software libs.

PR #1019, opened 2023-09-27
Fix -release repository check in release.py

j-rivero:
Fix the fail in the check for github repository, that now displays the error:
```bash
Error running command (git ls-remote -q --exit-code https://foo:foo@github.com/gazebo-release/gz-cmake-release).
stdout: 
stderr: remote: Support for password authentication was removed on August 13, 2021.
remote: Please see https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
fatal: Authentication failed for 'https://github.com/gazebo-release/gz-cmake-release/'
```

### DIFF
--- a/release.py
+++ b/release.py
@@ -183,12 +183,8 @@ def parse_args(argv):
     return args
 
 def get_release_repository_info(package):
-    # Do not use git@github method since it fails in non existant repositories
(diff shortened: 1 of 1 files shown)


scpeters (approved):
works for me

Issue #923, opened 2023-05-16
Reproduce the Jenkins CI linux build locally


j-rivero:
It should be possible trying to reproduce a Linux build on Jenkins in a local system in a non very difficult way. They mainly consist in a Docker enviroment + build.sh script and a custom way of running them in using `docker run` parameters. 

Roadmap:
 
 * [x] Export `Dockerfile` and `build.sh` from each Linux build https://github.com/gazebo-tooling/release-tools/pull/279
 * [ ] Fix hard-coded paths and values in Dockerfile and build.sh so a local system can run them without modifications. Goal is also to be able to run them using rocker and all the user developer tools via `--home`.
 * [ ] Use an easy way for running docker build + docker run with all the arguments setup correctly.
 * [ ] Find a quick way of setting the whole thing just by passing a Jenkins URL. Probably a `gzdev` small module.

j-rivero:
> Use an easy way for running docker build + docker run with all the arguments setup correctly.

docker build uses `$USER/$GUI/$UID` variables from active session. Can not find a way of using devcontainer.json file to include them.

Using docker compose it is also not fully clean since [UID/GUI seems not easy to deal with](https://github.com/compose-spec/compose-go/pull/299):
```yaml
services:
  jenkins-build:
    image: buildoo/jenkinsbuild
    build:
      context: .
      dockerfile: Dockerfile
      args:
        GID: ${GID}
        USERID: ${USERID}
        USER: ${USER}
``` 
Via:
```bash
USERID=$(id -u) GID=$(id -g) docker compose up
```

j-rivero:
While debugging an `-install-` type of job, I've been developing the compose configuration to something that is usable to reproduce the build locally for an X kind of job:

```yaml
services:
  jenkins-build:
    image: buildoo/jenkinsbuild
    build:
      context: .
      dockerfile: Dockerfile
      args:
        GID: ${GID}
        USERID: ${USERID}
        USER: ${USER}
  run-jenkins-build:
    image: buildoo/jenkinsbuild
    stdin_open: true 
    tty: true
    environment:
        - DISPLAY=unix${DISPLAY}
        - TERM=xterm-256color
        - GID=${GID}
        - USERID=${USERID}
        - USER=${USER}
    volumes:
        - /sys:/sys:ro
        - /tmp/.X11-unix:/tmp/.X11-unix:rw
        - /dev/log:/dev/log:ro
        - /run/log:/run/log:ro
[...]

PR #1030, opened 2023-10-11
Generate core dumps for tests that segfault

azeey:
This would help in debugging tests that crash on CI, but not locally.  The docker flags are based on https://ddanilov.me/how-to-configure-core-dump-in-docker-container

@j-rivero @nuclearsandwich is this a potential security risk? I didn't think it was since the tests are running inside docker and AFAIK the workspace is not accessible by users that are not logged in to Jenkins, but I wanted to confirm with y'all.

### DIFF
--- a/jenkins-scripts/docker/lib/docker_run.bash
+++ b/jenkins-scripts/docker/lib/docker_run.bash
@@ -12,6 +12,11 @@ mkdir -p ${PACKAGE_DIR}
 sudo rm -fr ${WORKSPACE}/build
 mkdir -p ${WORKSPACE}/build
 
+# Remove old core dumps to conserve space
+CORE_DUMPS_DIR="${WORKSPACE}/core_dumps"
+sudo rm -rf ${CORE_DUMPS_DIR}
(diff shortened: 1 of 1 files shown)


j-rivero:
> @j-rivero @nuclearsandwich is this a potential security risk? I didn't think it was since the tests are running inside docker and AFAIK the workspace is not accessible by users that are not logged in to Jenkins, but I wanted to confirm with y'all.

Discussed in the infra meeting. The public publication of the core dump file might be a security risk given the information dumped in these files. You are right about that we protect the workspace of Jenkins to be read only by admins, buildfarmers and github osrf/employees group (to be updated) so technically it could be safe. If other external contributors need access to them, they should be manually inspected first. Said all this, we believe that the best approach would be to have a reproducible build (see #923) but there is not ETA when that will be done.

Another thing to consider from the core dumps is that they are quite big, several Gbs usually. If we enable the generation of them automatically, we could have problems with disk space in nodes. How about introducing a parameter to activate the core dump exportation?

PR #1032, opened 2023-10-11
Retry brew update to fix errors about another active Homebrew process being active

azeey:
We're getting the "Another active Homebrew update process is already in progress" error very frequently now, so this is an attempt to fix it. ~It doesn't seem like a great approach since homebrew should be cleaning things up itself, but might be worth a try. At least, this is something others have [recommended](https://stackoverflow.com/a/40473139/283225). From the [logs](https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew-amd64/1656/console), `brew` is able to install things before `_homebrew_cleanup.bash` is called, so I figured removing the locks when that script is done might fix the issue.~ The fix retries running `brew update` several times until it succeeds.

### DIFF
--- a/jenkins-scripts/lib/_homebrew_base_setup.bash
+++ b/jenkins-scripts/lib/_homebrew_base_setup.bash
@@ -16,7 +16,15 @@ fi
 
 git -C $(${BREW_BINARY} --repo) fsck
 export HOMEBREW_UPDATE_TO_TAG=1
-${BREW_BINARY} update
+
+# There might be a background process that blocks `brew update`, so we try to
+# run it several times until it succeeds.
+# See https://github.com/Homebrew/brew/issues/1155
+brew_update_retry_count=0
+until ${BREW_BINARY} update || (( brew_update_retry_count++ > 6 ))
+do
+  sleep 10
+done
 # manually exclude a ruby warning that jenkins thinks is from clang
 # https://github.com/osrf/homebrew-simulation/issues/1343
 ${BREW_BINARY} install ${BREW_BASE_DEPENDCIES} \
--- a/jenkins-scripts/lib/_homebrew_cleanup.bash
+++ b/jenkins-scripts/lib/_homebrew_cleanup.bash
@@ -39,16 +39,6 @@ git stash && git clean -d -f
 ${BREW_BINARY} audit cmake || restore_brew
 popd 2> /dev/null
 
-# There might be a background process that blocks `brew update`, so we try to
-# run it several times until it succeeds.
-# See https://github.com/Homebrew/brew/issues/1155
-brew_update_retry_count=0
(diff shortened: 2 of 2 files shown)


scpeters:
there was a related issue https://github.com/Homebrew/brew/issues/13521 that was fixed in https://github.com/Homebrew/brew/pull/13586. Also, there was an earlier bug report https://github.com/Homebrew/brew/issues/1155 that mentioned that git may be auto-packing the repository in the background, which I believe was the motivation for the `brew fsck` calls added in https://github.com/gazebo-tooling/release-tools/commit/f5e9ef9ad6bef555ca4e30d04d4c10f3ff59bc4e.

Also, Mike McQuaid was clearly 👎 on blindly removing locks in https://github.com/Homebrew/brew/issues/1155#issuecomment-249796351.

If we want a workaround, perhaps we can sleep and retry `brew update` a few times to see how long it takes to release the other process?

azeey:
Okay, using retries instead in https://github.com/gazebo-tooling/release-tools/pull/1032/commits/569680033fd77dc77a2d5a29903ffe8709b7a948

azeey:
Test builds:
[Build Status](https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew-amd64/1663/)
[Build Status](https://build.osrfoundation.org/job/ignition_transport-ci-pr_any-homebrew-amd64/1662/)
[Build Status](https://build.osrfoundation.org/job/sdformat-ci-pr_any-homebrew-amd64/5032/)

scpeters:
I looked at the following recent failure, and I think it failed at the following `brew update` invocation:

* https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/lib/_homebrew_base_setup.bash#L19

azeey:
> I looked at the following recent failure, and I think it failed at the following `brew update` invocation:
> 
> * https://build.osrfoundation.org/view/os_homebrew/job/ignition_gazebo6-install_bottle-homebrew-amd64/742/console
> * https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/lib/_homebrew_base_setup.bash#L19

Ah, you're right. If the problem is potentially do to the `git fsck`, it makes sense to put the retries there. Done in https://github.com/gazebo-tooling/release-tools/pull/1032/commits/33c09517ef63c1140be2fff04d4f78d3c2e43b60 and fixed a redundancy in calling `brew update`.

scpeters:
testing

[Build Status](https://build.osrfoundation.org/view/os_homebrew/job/ignition_gazebo6-install_bottle-homebrew-amd64/744/) https://build.osrfoundation.org/view/os_homebrew/job/ignition_gazebo6-install_bottle-homebrew-amd64/744/

scpeters (approved):

scpeters:
let's try it!

PR #1047, opened 2023-10-25
Change daily Linux ci jobs (-ci-${branch}) to be based on gz-collections.yaml

j-rivero:
Remove the protection in `gazebo_libs` after Harmonic jobs were ready to generate the -ci-${branch}- jobs for Linux. Remove the corresponding code in ignition.dsl and sdformat.dsl. I would expect some changes in the branch names now that we are consistent with the yaml spec but the number of jobs and the testing done should be equivalent.

@Crola1702 @Blast545 For the buildfarmer board: we probably want to use the new log of jobs generated that help to remove the hardcoded values and would be auto-updated if something change at any point.

"+0 -172 lines" :grinning:

PR #1050, opened 2023-10-27
Update outdated-job-runner job list

Crola1702:
After https://github.com/gazebo-tooling/release-tools/pull/1047, naming on garden, fortress, and citadel changed for Linux

This PR updates the job list for periodic runs


Blast545 (approved):
Do we have any way to check this script before merging pull requests changes to it? Or compare it against our internal common.py?

Something to think about once you are back.

j-rivero:
Side note: it would be great to use the output of the own job generation (see https://github.com/gazebo-tooling/release-tools/pull/985) to somehow feed this list. Not for this PR but a nice future work.

claraberendsen:
> Side note: it would be great to use the output of the own job generation (see #985) to somehow feed this list. Not for this PR but a nice future work.

Agreed that it would be nice to connect it to the changes on the dsl jobs. I will add it to our backlog for future work.

claraberendsen (approved):
This LGTM in terms of getting us up to date with the current approach. However for these names I think we should push to have one source of truth that we fetched this information from rather than manually updating each time, given that this affects us a couple of number of times in the year.

claraberendsen:
@Blast545 feel free to merge this as is for now, and let's add a new tasks on how we can keep this list up-to-date with the changes in the jobs.

PR #1054, opened 2023-11-02
Move brew ci-pr_any- to gazebo_libs

j-rivero:
Note: easier to check without whitespaces https://github.com/gazebo-tooling/release-tools/pull/1054/files?w=1

Change brew ci-pr_any jobs to be created by gazebo_libs.dsl using `gz-collections.yaml`.

 Steps:

 * Define a new CI Configuration in gz-collections.yaml for Brew edae775d013520341514a9028837adae48a08010
 * Implement the new jobs in `gazebo_libs.dsl` (no whitespaces enabled) https://github.com/gazebo-tooling/release-tools/commit/c9e5af8fcbb5655cf128b2a75e8d749cb5045a14?w=1
 * Use the trick of supporting all branches (not only stable targets) until #1044 is solved 0b1e7ec8f08f8819c0db1d612042d193da313a02
 * Modify existing jobs to return always true before we remove them in some days c7b164e687b35140a50c8370837b938682801e6d

PR #1060, opened 2023-11-07
Fix brew -ci-pr_any-: remove workarounds in project-default bash file

j-rivero:
After #1054 the brew builds are broken with problem to execute the detection of the major versions. Problem seems to be in the bash code of `jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash` that was modifying the value passed as the relative path for workaround to force the `ign-` prefix injection. We should not need that anymore.

Failing:
gz-cmake [Build Status](https://build.osrfoundation.org/job/gz_cmake-ci-pr_any-homebrew-amd64/1/)

This PR:
gz-cmake [Build Status](https://build.osrfoundation.org/job/gz_cmake-ci-pr_any-homebrew-amd64/2/)


j-rivero:
Merging fast to fix broken CI.

PR #1075, opened 2023-11-17
brew: use HOMEBREW_PREFIX in PYTHONPATH

scpeters:
The `PYTHONPATH` currently references `/usr/local`, but it is cleaner to use the `HOMEBREW_PREFIX` variable to match how other variables are defined. This will be most relevant when we have mac CI machines with ARM CPUs.


scpeters:
testing:

* [Build Status](https://build.osrfoundation.org/job/sdformat-ci-sdf14-homebrew-amd64/2/) https://build.osrfoundation.org/job/sdformat-ci-sdf14-homebrew-amd64/2/

This job has a pre-existing cmake warning, but I want to confirm that the python tests still pass with this branch.

scpeters:
I'm not sure this line is even right...

Issue #1010, opened 2023-09-11
Tracker: job migration to gazebo_libs.dsl (yaml)


j-rivero:
The issue stores the information on the migration of the different CI jobs from being hardcoded in DSL and generated by [`ignition.dsl`](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/ignition.dsl) / [`ignition_collection.dsl`](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/ignition_collection.dsl) and [`sdformat.dsl`](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/sdformat.dsl) to be stored in a permanent configuration place: [`gz-collections.yaml`](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/gz-collections.yaml) (still open for future changes but a yaml file by now) and generated via [`gazebo_libs.dsl`](https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/gazebo_libs.dsl)

* Yaml
   - Distribution information
     - [ ] Gazebo Citadel
     - [ ] Gazebo Fortress
     - [x] Gazebo Harmonic
  - Type of jobs
    - Linux 
      - [ ] PR testing `gz_${lib}-ci-pr_any-ubuntu_${distro}` jobs
        - [x] Harmonic/Jammy
      - [ ] Branch main CI:  `gz_${lib}-ci-gz-${branch}-${distro}` jobs
        - [x] Harmonic/Jammy

* DSL (`ignition.dsl` / `sdformat.dsl`)
  - Type of jobs
    - Linux 
      - PR testing `ign_${lib}-ci-pr_any-ubuntu_auto` jobs 
        - Included Harmonic on Focal
      - ABI checker testing `ign_${lib}-abichecker-any_to_any-ubuntu_auto` jobs
         - Included Harmonic on Focal       
      - Install testing `gz/ign_${lib}-install-pkg-${distro}` jobs
        - Included Harmonic on Focal
      - CI Asan `ignition_${lib}-ci_asan-ign/gz-${branch}-${distro}` 
         - Included Harmonic on Focal 
      - Debbuilders
  - Brew (all)
  - Windows (all)

scpeters:
[...]

PR #1079, opened 2023-11-20
Jenkins views: fix some job names

scpeters:
Changes some `ignition_gazebo` to `gz_sim` and other `ignition_` to `gz_`.

UPDATE:

I've added many more changes since that first commit. There are still some issues:

* Some windows CI builds are missing for Fortress
* The [_dsl_gazebo_libs](https://build.osrfoundation.org/job/_dsl_gazebo_libs/) job doesn't seem to generate any focal builds for the citadel branches


j-rivero:
One of the reasons to implement the issue #1010 is to be able to generate all the collections set of jobs programmatically and do not hardcode values anymore. The implement still lacks of the `-install-` jobs to be able to replace the views in this PR but can provide all the ci-branch related job names in a log file together with the type of job and the collection. This is the current log https://gist.github.com/j-rivero/7c3468bae2e9ee39497131715c6add82.

It is easy to alpha-sort + grep that generated log file so two options:
  * Just copy/paste the alpha-order of all ci-branch and move the -install- jobs after.  It will change current order and make things more difficult to human-read but would be great for completeness.
  * We leave the PR as it is and wait unlit the -install- jobs generation to replace it completely.

Whatever you prefer @scpeters

scpeters:
> One of the reasons to implement the issue #1010 is to be able to generate all the collections set of jobs programmatically and do not hardcode values anymore.

ooh, I love this! 💯 

> The implement still lacks of the `-install-` jobs to be able to replace the views in this PR but can provide all the ci-branch related job names in a log file together with the type of job and the collection. This is the current log https://gist.github.com/j-rivero/7c3468bae2e9ee39497131715c6add82.
> 
> It is easy to alpha-sort + grep that generated log file so two options:
> 
> * Just copy/paste the alpha-order of all ci-branch and move the -install- jobs after.  It will change current order and make things more difficult to human-read but would be great for completeness.
> * We leave the PR as it is and wait unlit the -install- jobs generation to replace it completely.

I think I will paste the values from your gist into this branch just to get working views again, and then we can deprecate the explicit list for the auto-generated views

scpeters:
to the best of my ability, I've manually updated the job names to match the [gist](https://gist.github.com/j-rivero/7c3468bae2e9ee39497131715c6add82) that @j-rivero listed in https://github.com/gazebo-tooling/release-tools/pull/1079#issuecomment-1821544221. So this currently matches the names generated by our DSL jobs. We can discuss what we would like them to be named and work on auto-generating the views, but I think this PR could be merged as a short-term fix for the views

j-rivero:
> We can discuss what we would like them to be named and work on auto-generating the views, but I think this PR could be merged as a short-term fix for the views

Agree. LGTM

PR #1080, opened 2023-11-21
Use core.symlinks in gazebodistro for windows checkout

j-rivero:
Use `-c core.symlinks=true` in git clone to correctly grab all the `ign_` in gazebodistro.

Tested: [Build Status](https://build.osrfoundation.org/job/ign_gazebo-3-win/)


j-rivero:
Although might be the right choice for the clone, it is not going to work due to problems with GAZEBODISTRO file and the way of overriding the directories from the vcs import and the copy of the checkout directory.

PR #1084, opened 2023-11-28
Add jobs.txt first version

Crola1702:
Inside the buildfarmer workflow, we're tracking a specific set of jobs that are generated by release-tools, this list is called `jobs.txt`.

This should be the first version that match the jobs generated in the version 70a19a1b.

> **Note:**
> This `jobs.txt` version doesn't have most of the changes that are proposed right now.


Crola1702:
Maybe as a suggestion, we could change the name of the file to be something easier to understand by an external contribution like `buildfarm-tracked-jobs.txt`.

WDYT @j-rivero?

j-rivero:
Some comments as discussed with the buildfarmer team:
 * My personal feeling is that keeping a file that can be easily generated from the source code under the control of the version control system has a good number of difficulties.
 * With this PR the requirement is that anyone that make changes to DSL code needs to generate the `jobs.txt` file locally and commit the changes together with the code changes. This look to my eyes error prone and the maintenance cost of being sure that the file is updated when someone send a PR and interact to explain the people how do it won't be trivial.

My proposal is to hold this until we have mechanism for updating the `jobs.txt` automatically.

PR #1086, opened 2023-11-30
Clean up DSL CI by using GitHub actions paths

j-rivero:
Modify the GitHub actions workflow to rely on `paths:` for triggering the CI instead of having a custom logic of detecting changes on dsl directory.


j-rivero:
Going to test it in the next dashboard PR.

PR #1087, opened 2023-11-30
Update dashboard for changes in Citadel

j-rivero:
Corresponding to changes in #1083


j-rivero:
I've run the DSL generation and obtain the new https://build.osrfoundation.org/view/ign-citadel/. Also triggered new builds in all jobs without a previous run.

PR #1089, opened 2023-11-30
Automatically update jobs.txt log on pull request changes in DSL

j-rivero:
Following up https://github.com/gazebo-tooling/release-tools/pull/1084#issuecomment-1831944935 .

The PR implements a mechanism to update jobs generated file on pull request updates. It uses a third party action https://github.com/marketplace/actions/git-auto-commit to implement the commit action and I set the author of the commits to `GitHub Bot in release-tools <github-actions-bot@local>`.

A good bunch of code is removed from existing action since the log diff is now obtained in the GitHub PR web interface.

The PR needs to import the current job generated list generated from master at the time of merging.


Blast545 (approved):
Pre-review: LGTM

.github/workflows/ci.yaml
```diff
@@ -6,61 +6,52 @@ on:
...
-          # export files for later diff
-          mkdir /tmp/pr_xml_configuration && mkdir /tmp/pr_log_generated/
-          # sort jobs.txt
-          sort jobs.txt -o jobs.txt
-          mv *.xml /tmp/pr_xml_configuration/
-          mv *.txt /tmp/pr_log_generated/
+          find logs/* -exec sort {} -o {} \;
+          popd
+      - name: Update and commit logs generated
+        uses: stefanzweifel/git-auto-commit-action@v5
```
nuclearsandwich, line 35: By default this action adds the author whose commit triggered the action as a commit co-author. Does the configuration below preserve that extra information?

.github/workflows/ci.yaml
```diff
@@ -6,61 +6,52 @@ on:
       - 'jenkins-scripts/dsl/**'
 
 jobs:
-  dsl_ci:
+  xml_generation:
     runs-on: ubuntu-latest
-    name: Diff for DSL code
+    name: Generate XML config from DSL
```
nuclearsandwich, line 11: According to the upstream docs you need to give the default token scope to change repository contents.
```suggestion
    name: Generate XML config from DSL
    permissions:
      contents: write
```

.github/workflows/ci.yaml
```diff
@@ -6,61 +6,52 @@ on:
       - 'jenkins-scripts/dsl/**'
```
nuclearsandwich, line 6:
```suggestion
      - 'jenkins-scripts/dsl/**'
```

Does it make sense to update the triggering mechanisms here so that updates only occur on the main branch or make any other affordances to prevent abuse of these workflows?

I think the risk is limited to the logs directory without very obvious interference with the workflows and since this is a metadata update not a production deployment we don't need to lock this down as tightly. It is also nice to have the change auto-committed _before_ the PR is merged rather than after, which is what my alternative strategy requires.

nuclearsandwich (commented):
I had a couple of questions and noticed a possible permissions issue.

.github/workflows/ci.yaml
j-rivero: The change looks just correct to me. I was investigating why the action was working on my GitHub fork and found the repository configuration:

Screenshot from 2023-12-06 16-42-15

Might be a good moment to change this.

.github/workflows/ci.yaml
j-rivero: It does not. Default behavior is nicer indeed [e973aeb](https://github.com/gazebo-tooling/release-tools/pull/1089/commits/e973aebb7923f3d5aad746928349d186314425e3)

.github/workflows/ci.yaml
nuclearsandwich-ai: > This will cover direct commits to `master` which would be a good thing to have. However, as you mentioned, showing the diff on the PR before it is merged is really useful.

Yeah this was the alternative I was trying to describe. I think that for this type of change, having it performed on the PR branch before merge is preferred. So it's good as is.

.github/workflows/ci.yaml
j-rivero: changed.

j-rivero:
> The one outstanding action is a change of settings for the repository which will happen outside the context of this PR.

Done.

nuclearsandwich:
Just stating for the public record that @nuclearsandwich-ai is an account associated with my current dayjob work and was used for this review by mistake since I opened the link in the wrong browser session.

PR #1091, opened 2023-12-05
Improve outdated job feching and success strategy

Crola1702:
With the capabilities that https://github.com/gazebo-tooling/release-tools/pull/1084 provides. We now don't need to manually define the jobs to run, but instead, we could get them dynamically with `jobs.txt` file.

Also, the job will now be marked as unstable if it tries to run a disabled project (e.g., [ign_cmake-2-win](https://build.osrfoundation.org/job/ign_cmake-2-win/) is disabled).

And the job will fail if the project doesn't exist or if it doesn't find the job list.

Pending task:
- [x] Get real tracked jobs list URL


jenkins-scripts/tools/outdated-job-runner.groovy
```diff
@@ -52,181 +53,33 @@ def runJobsInAvailableNodes(LinkedHashMap outdatedJobs) {
...
-    'sdformat-ci-sdformat14-homebrew-amd64',
-    'sdformat-ci-sdformat9-bionic-amd64',
-    'sdformat-ci-sdformat9-homebrew-amd64',
-    'sdformat-sdf-12-win',
-    'sdformat-sdf-13-win',
-    'sdformat-sdf-14-win',
-    'sdformat-sdf-9-win',
-];
-def trackedJobs = new LinkedHashSet(trackedJobsList);
+def JOBS_URL = 'https://raw.githubusercontent.com/Crola1702/release-tools/master/jenkins-scripts/dsl/jobs.txt'
```
j-rivero, line 69: Please check #1089 to see the final path and final name of the log file generated.

j-rivero (approved):
I've removed a println and integrate the return code in [29e9996](https://github.com/gazebo-tooling/release-tools/pull/1091/commits/29e9996023e1c12ead9a0fcc76551d7e73df418d). 

Otherwise looks good.

PR #1094, opened 2023-12-13
Do not exclude gz-common from focal ci_configs

azeey:
ign-common4 (Fortress) is configured to run on `focal`, but we were excluding the abichecker from the `focal` ci_configs.

https://build.osrfoundation.org/job/gz_common-abichecker-any_to_any-ubuntu-jammy-amd64/ only triggers on `gz-common5`.

### DIFF
--- a/jenkins-scripts/dsl/gz-collections.yaml
+++ b/jenkins-scripts/dsl/gz-collections.yaml
@@ -466,7 +466,6 @@ ci_configs:
         - gz-cmake
         - gz-tools
         # These have a Jammy counterpart in Harmonic
-        - gz-common
         - gz-math
         - gz-plugin
         - gz-utils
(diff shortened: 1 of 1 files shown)


azeey:
> In fact, since we moved Citadel from Bionic to Focal we also need to remove the other exclusions:
> 
> ```yaml
>         - gz-math
>         - gz-plugin
>         - gz-utils
> ```
> 
> This is going to have the side effect of making the branches shared by Garden (ci_config is focal) and Harmonic(ci_config is jammy) to duplicate the abichecker jobs. I'll open an issue and think the best way of dealing with this.

Thanks for the fix!

azeey:
Can we merge this  before we find a solution to #1095 to unblock https://github.com/gazebosim/gz-common/pull/559?

j-rivero:
> Can we merge this before we find a solution to #1095 to unblock [gazebosim/gz-common#559](https://github.com/gazebosim/gz-common/pull/559)?

sure!

PR #1097, opened 2023-12-14
Make Citadel to use custom ciconfig to exclude gz-sim/gz-launch

j-rivero:
Need to exclude gz-sim and gz-launch from Citadel. To avoid adding harcoded values in the DSL, the declarative approach is to create a new ciconfig for Citadel, just by duplicating the win configuration and make the exclusions for the libraries.

### DIFF
--- a/jenkins-scripts/dsl/gz-collections.yaml
+++ b/jenkins-scripts/dsl/gz-collections.yaml
@@ -96,7 +96,7 @@ collections:
       configs:
         - focal
         - brew
-        - win
+        - win_citadel
     packaging:
       configs:
         - focal
@@ -532,7 +532,6 @@ ci_configs:
     requirements:
     exclude:
       all:
-        - gz-citadel
         - gz-fortress
         - gz-garden
         - gz-harmonic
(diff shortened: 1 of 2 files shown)


jenkins-scripts/dsl/gz-collections.yaml
```diff
@@ -96,7 +96,7 @@ collections:
       configs:
         - focal
         - brew
-        - win
```
Blast545, line 99: IIUC this PR this line has to be kept(? Then you would have focal, brew, win and win_citadel configs

Blast545 (commented):
Small question jic, otherwise LGTM

jenkins-scripts/dsl/gz-collections.yaml
j-rivero: Note that this `config` section [is defining the configurations for Citadel](https://github.com/gazebo-tooling/release-tools/pull/1097/files#diff-f69908edf9a24c4bb5dca1219f6d179a15eb2e65739e410c65df6d343d3d2b63L29-L96) so win is replaced by win_citadel.

You are right about that now we have 4 configurations available, with `win`still [defined generally](https://github.com/gazebo-tooling/release-tools/pull/1097/files#diff-f69908edf9a24c4bb5dca1219f6d179a15eb2e65739e410c65df6d343d3d2b63L526) in the `ci_configs` section and used for Fortress, Garden, Harmonic, Ionic and __upcoming__.

Blast545: Oh, I see, thanks for the explanation

Blast545 (approved):

PR #1100, opened 2023-12-22
ros_gz wrappers: include ROS package in conflicts when adding a second conflict

j-rivero:
When creating the ros_gz unofficial packages, if `GZ_RELEASE_TO_CONFLICT` is set the conflict on the original ROS package is lost. For example: releasing Harmonic packages for Humble that conflict with the unofficial Garden packages make the packages not to conflict with `ros-humble-ros-gz`.

The PR preserves the original conflict and add a new one when using `GZ_RELEASE_TO_CONFLICT`


j-rivero:
Going to merge this one not to lost it in the next use.

PR #1110, opened 2024-01-26
Homebrew CI: set Python3_EXECUTABLE in cmake

scpeters:
If the formula of the package being tested defines a `python_cmake_arg` method, then set `Python3_EXECUTABLE` when invoking cmake. Also append to `CMAKE_ARGS` instead of overwriting it in `qwt-qt5` logic.

Pairs with https://github.com/osrf/homebrew-simulation/pull/2543.

### DIFF
--- a/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
@@ -147,6 +147,10 @@ fi
 if brew ruby -e "exit ! '${PROJECT_FORMULA}'.f.recursive_dependencies.map(&:name).keep_if { |d| d == 'icu4c' }.empty?"; then
   export LIBRARY_PATH=${LIBRARY_PATH}:/usr/local/opt/icu4c/lib
 fi
+# set Python3_EXECUTABLE if this homebrew formula defines the python_cmake_arg method
+if brew ruby -e "exit '${PROJECT_FORMULA}'.f.respond_to?(:python_cmake_arg)"; then
+  CMAKE_ARGS="${CMAKE_ARGS} -DPython3_EXECUTABLE=$(which python3)"
+fi
 
 # if we are using dart@6.10.0 (custom OR port), need to add dartsim library path since it is keg-only
 if brew ruby -e "exit ! '${PROJECT_FORMULA}'.f.recursive_dependencies.map(&:name).keep_if { |d| d == 'osrf/simulation/dartsim@6.10.0' }.empty?"; then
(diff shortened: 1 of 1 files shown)


scpeters:
Testing this branch along with https://github.com/osrf/homebrew-simulation/pull/2543 (via ci_matching_branch/) in https://github.com/gazebosim/gz-sim/pull/2298:

* [Build Status](https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/240/) https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/240/

jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
```diff
@@ -147,6 +147,10 @@ fi
 if brew ruby -e "exit ! '${PROJECT_FORMULA}'.f.recursive_dependencies.map(&:name).keep_if { |d| d == 'icu4c' }.empty?"; then
   export LIBRARY_PATH=${LIBRARY_PATH}:/usr/local/opt/icu4c/lib
 fi
+# set Python3_EXECUTABLE if this homebrew formula defines the python_cmake_arg method
+if brew ruby -e "exit '${PROJECT_FORMULA}'.f.respond_to?(:python_cmake_arg)"; then
+  CMAKE_ARGS="${CMAKE_ARGS} -DPython3_EXECUTABLE=$(which python3)"
```
azeey, line 152: Would it be possible to get the value of `CMAKE_ARGS` from the `python_cmake_arg` function so that the python logic is defined in one place?

scpeters: I spent some time trying to do this and ran into some roadblocks with how `PATH` is set for the `brew ruby` tool. I tried the following, and I think it's more elegant,

~~~
CMAKE_ARGS="${CMAKE_ARGS} $(brew ruby -e 'puts ${PROJECT_FORMULA}.f.python_cmake_arg')"
~~~

but it doesn't work:

~~~
# python3 is found first in /usr/local/bin
$ which python3
/usr/local/bin/python3
# but `brew ruby` finds python3 from /usr/bin
$ brew ruby -e "puts 'gz-msgs10'.f.python_cmake_arg"
-DPython3_EXECUTABLE=/usr/bin/python3
# `brew ruby` uses a controlled environment by default
$ brew ruby -e "puts ENV['PATH']"
/usr/local/Homebrew/Library/Homebrew/shims/shared:/usr/bin:/bin:/usr/sbin:/sbin
~~~

An alternative is to manually prepend `"#{HOMEBREW_PREFIX}/bin"` to the `PATH` in the `brew ruby` call. This feels clunky to me, which is why I didn't initially suggest it, but we can use it if you like it better. I will also follow up with brew folks about a cleaner way to do this:

~~~
brew ruby -e 'ENV["PATH"] = PATH.new(ENV.fetch("PATH")).insert(1, HOMEBREW_PREFIX/"bin").to_s; puts "gz-msgs10".f.python_cmake_arg'
~~~

scpeters: I'm not expecting any big changes in the way `brew ruby` works, so I think the workarounds I've listed here are what we can choose between

azeey: The original approach seems better to me.

azeey (approved):

PR #1112, opened 2024-02-05
Removes apt-get installed packages from Dockerfile

claraberendsen:
### Description
There were some issues with unattended upgrades making builds fail ([See](https://build.osrfoundation.org/job/gz_sim-abichecker-any_to_any-ubuntu-jammy-amd64/120)). After some research it seems this is due to release-tools installing packages that are already provisioned when the image for the agents is built. This PR removes all dependencies that are installed through this mechanism.

### DIFF
--- a/jenkins-scripts/docker/lib/boilerplate_prepare.sh
+++ b/jenkins-scripts/docker/lib/boilerplate_prepare.sh
@@ -144,7 +144,7 @@ output_dir=$WORKSPACE/output
 work_dir=$WORKSPACE/work
 
 # TODO: Check for docker package
-NEEDED_HOST_PACKAGES="git python-setuptools python-psutil qemu-user-static gpgv squid-deb-proxy bc"
+NEEDED_HOST_PACKAGES="git python-setuptools python-psutil gpgv squid-deb-proxy bc"
 # python-argparse is integrated in libpython2.7-stdlib since raring
 # Check for precise in the HOST system (not valid DISTRO variable)
 if [[ $(lsb_release -sr | cut -c 1-5) == '12.04' ]]; then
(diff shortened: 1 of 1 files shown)


Blast545:
Just for the sake of double checking it doesn't break the build where the problem was found, can you run a job in https://build.osrfoundation.org/job/gz_sim-abichecker-any_to_any-ubuntu-jammy-amd64/build?delay=0sec with this fix branch?

Blast545 (approved):

j-rivero:
> Just for the sake of double checking it doesn't break the build where the problem was found, can you run a job in https://build.osrfoundation.org/job/gz_sim-abichecker-any_to_any-ubuntu-jammy-amd64/build?delay=0sec with this fix branch?

Clarifying: to avoid the problem of having two conflicting package installation processes we should remove all of the packages in this call to be able to get rid of it. To do so, the provisioning should be patches to install all the packages installed by `NEEDED_HOST_PACKAGES`. I did it in https://github.com/osrf/osrf_jenkins_agent/pull/43.

To completely kill the apt-get install, we would need to merge the PR in provisioning and re-generate new AIMs.

j-rivero (approved):

claraberendsen:
> To completely kill the apt-get install, we would need to merge the PR in provisioning and re-generate new AIMs.
@j-rivero I'm agree on unifying everything on provisioning. I will update this PR to remove the apt-get call completely and proceed to review and deploy the related one you opened before merging this one.

claraberendsen:
@j-rivero re-requesting a review here since we changed approach.

j-rivero (approved):

j-rivero:
> @j-rivero re-requesting a review here since we changed approach.

Can not merge before we re-generate the AMIs of the agents.

PR #1113, opened 2024-02-08
Change nightly labeler to run every 20min all day

j-rivero:
We want to be sure that there is a nightly node labeled always, not just during the nightly creation or generation. The PR uses a cron expression that tries to be sure that we still run it at 9:00, 10 minutes before the nightly creation but also add the rest of the day.


Crola1702 (approved):
LGTM

PR #1119, opened 2024-02-23
brew CI: pip install with --break-system-packages

scpeters:
Try this until we switch to a venv.

This should allow `pytest` to be installed, which provides better feedback to jenkins about python test results.

Testing:

* [Build Status](https://build.osrfoundation.org/job/gz_math-ci-ign-math6-homebrew-amd64/25/) https://build.osrfoundation.org/job/gz_math-ci-ign-math6-homebrew-amd64/25/
* [Build Status](https://build.osrfoundation.org/job/sdformat-ci-sdf14-homebrew-amd64/25/) https://build.osrfoundation.org/job/sdformat-ci-sdf14-homebrew-amd64/25/


scpeters:
I expected the sdf14 build to be unstable since the homebrew formulae need to be updated to use `python@3.12`, and with `pytest` installed, it is properly showing the python test results, so I think the CI results are as expected

azeey (approved):

PR #1127, opened 2024-04-05
brew: pip install protobuf in CI

azeey:
gz-transport Python tests are failing because Python bindings were removed in the latest update of protobuf homebrew bottle. It seems like the [recommended](https://github.com/protocolbuffers/protobuf/tree/main/python#installation) way to install the Python bindings is through `pip`. Since this will be needed for anything above gz-transport, I thought we should just add it here to `PIP_PACKAGES_NEEDED` as a global dependency.

### DIFF
--- a/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
@@ -78,6 +78,9 @@ brew install $(brew deps --1 --include-build ${PROJECT_FORMULA})
 # pytest is needed to run python tests with junit xml output
 PIP_PACKAGES_NEEDED="${PIP_PACKAGES_NEEDED} pytest"
 
+# Add protobuf since the homebrew protobuf bottle dropped support for python bindings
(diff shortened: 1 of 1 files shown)


scpeters:
Related:

* https://github.com/Homebrew/homebrew-core/pull/166008
* https://github.com/Homebrew/homebrew-core/pull/166096
* https://github.com/Homebrew/homebrew-core/pull/166008#issuecomment-2002022287
* https://github.com/Homebrew/homebrew-core/pull/166008#issuecomment-2005118457

I will experiment with putting protobuf into formula `resource` blocks, but I'm guessing it won't work for users who just want to run `python3` and import and use our bindings

scpeters:
testing:

* [Build Status](https://build.osrfoundation.org/view/gz-harmonic/job/gz_msgs-ci-gz-msgs10-homebrew-amd64/41/) https://build.osrfoundation.org/view/gz-harmonic/job/gz_msgs-ci-gz-msgs10-homebrew-amd64/41/

scpeters (approved):

scpeters:
I think we're going to need to add `pip install protobuf` to our macOS installation instructions

PR #1128, opened 2024-04-09
Update and fix gzdev jobs in Jenkins

j-rivero:
Current gzdev jobs are hanging on the buildfarm. The PR updates a good bunch of the CI code in different aspects:

 * Modernize DSL distributions
 * Update and fix the docker-in-docker support until the point of being able to run rocker
 * Remove old docker install code

Still a work in progress is to fix the real CI done on gzdev docker module, it will probably require upstream code changes. I've commented out the support by now. Anyway, the changes in docker-in-docker and the fix for the handing code should worth the pull request.


j-rivero:
Tested it [Build Status](https://build.osrfoundation.org/job/gzdev-ci-default-jammy-amd64/)

PR #1130, opened 2024-04-11
Inject test output into junit files generated by gtest

azeey:
Fixes https://github.com/gazebosim/gz-cmake/issues/240

Example: https://build.osrfoundation.org/view/asan/job/gz_common-ci_asan-gz-common5-jammy-amd64/32/testReport/(root)/INTEGRATION_encoder_timing/test_ran/


j-rivero (changes requested):
Wow, this is great!  Needs to make some tweaks. My review of the script is in https://github.com/gazebo-tooling/release-tools/pull/1131, 3 small fixes/improvements.

I'm working on making the CI to test if the ctest version is recent enough to have the `--output-junit` (focal does not have it) and to generalize the support to we can include it on Mac.

azeey:
From my local tests, if ctest doesn't have `--output-junit`, it just simply ignores the flag and won't generate anything. The Python script will just not run in that case since we check for the output .xml file before running it.

j-rivero:
> From my local tests, if ctest doesn't have --output-junit, it just simply ignores the flag and won't generate anything. The Python script will just not run in that case since we check for the output .xml file before running it.

that is totally true. I was checking first with `cmake` instead of `ctest` and it fails hard if the argument does not exist.

j-rivero:
@Crola1702 @Blast545 @claraberendsen we are ready to merge this one. The side effect of it is that we can see new failures raised to the jenkins builds results as a consequence of raising the errors from CTest to the google-test files that are the ones scanned by Jenkins. I don't think that we have ASAN builds under the scope of the standard buildfarm report which are the ones mostly affected but some others could appear. New errors are probably valid and need to be triggered.

j-rivero (approved):
Good to go.

PR #1131, opened 2024-04-12
cmake_to_gtest_junit_output.py improvements (PR #1130)

j-rivero:
Three small changes for the #1130 script. Commits are self-descriptive.


jenkins-scripts/tools/cmake_to_gtest_junit_output.py
```diff
@@ -22,14 +22,17 @@
 
 def main(argv=sys.argv[1:]):
     parser = argparse.ArgumentParser(
-        description="Parse system output from junit file generated by CMake and inject them into junit files generated by gtest."
+        description="Parse system output from junit file generated by CTest"
+                    "and inject it into existing junit files generated "
+                    "by gtest. Useful in some cases like ASAN reports."
```
azeey, line 27: I think this is also useful when tests segault or timeout which means gtest doesn't get a chance to generate junit files. The Jenkins test results page for this types of tests only shows an error stating that the junit file was not found. We have to then go through the console logs to see the actual test output. See https://build.osrfoundation.org/view/gz-harmonic/job/gz_common-ci_asan-gz-common5-jammy-amd64/36/testReport/(root)/INTEGRATION_encoder_timing/test_ran/

azeey (approved):
Thanks for the fixes!!

jenkins-scripts/tools/cmake_to_gtest_junit_output.py
j-rivero: It is an amazing change! Want to see that in action. I'm also trying to make Jenkins to consume the generated file directly and see how it presents the test report.

azeey: I tried directly using the cmake junit output : https://build.osrfoundation.org/job/gz_utils-ci-pr_any-jammy-amd64/81/testReport/(root)/. It's a bit redundant.

j-rivero: > I tried directly using the cmake junit output : https://build.osrfoundation.org/job/gz_utils-ci-pr_any-jammy-amd64/81/testReport/(root)/. It's a bit redundant.

That is probably because the jenkins configuration is set to `$workspace/build/test_result/*.xml` (getting both gtest and ctest). I tried setting it up to just `$workspace/build/test_result/cmake_to_gtest_junit_output.xml` in https://build.osrfoundation.org/job/_test_gz_fuel_tools-ci_asan-gz-fuel-tools9-jammy-amd64/6/testReport/(root)/. You can compare:

  * Only new ctest junit https://build.osrfoundation.org/job/_test_gz_fuel_tools-ci_asan-gz-fuel-tools9-jammy-amd64/6/testReport/(root)/
  * Current gtest: https://build.osrfoundation.org/job/_test_gz_fuel_tools-ci_asan-gz-fuel-tools9-jammy-amd64/2/testReport/(root)/

By opening each test you can see that gtest consider each of the individual tests defined in source as independent while the ctest consider the whole file as a single test.

azeey: Yeah, the `ctest` output junit is not great for that aspect. We talked about using `gtest_discover_tests` in the [past](https://github.com/gazebosim/gz-cmake/issues/240#issuecomment-1523631274), but that has its own drawbacks. For now, I'm happy with just getting the console output of the test attached to each test result in Jenkins.

PR #1132, opened 2024-04-15
Make test to fail if problems where found in CTest results (for #1130)

j-rivero:
While testing #1130 using gz-fuel-tools9 ASAN branch just noticed that although we are injecting missing context into some tests, in the case of finding ASAN failures, the context is indeed injected but the failure state is not be injected so the tests are still passing all of them even if the status was "fail" in CTest.

With this PR I'm using the field "errors" reserved for non-CI strictly failing reasons to inject the errors from CTest into the Gtest testsuite. The PR also adds a bit of output about what we are doing and where.

See it working in https://build.osrfoundation.org/job/_test_gz_fuel_tools-ci_asan-gz-fuel-tools9-jammy-amd64/17/

### DIFF
--- a/jenkins-scripts/tools/cmake_to_gtest_junit_output.py
+++ b/jenkins-scripts/tools/cmake_to_gtest_junit_output.py
@@ -48,12 +48,25 @@ def main(argv=sys.argv[1:]):
             args.gtest_results_dir, testcase.attrib["name"] + ".xml"
         )
         system_output = testcase.find(SYSTEM_OUT_TAG)
+        testcase_failures = 0
         assert system_output is not None
         try:
+            print(f"Processing {results_file_name}")
             results_doc = ET.parse(results_file_name)
             should_write_to_file = False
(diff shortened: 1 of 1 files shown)


j-rivero:
Tested also here, good results for gz-transport asan https://build.osrfoundation.org/view/gz-ionic/job/gz_transport-ci_asan-main-jammy-amd64/32/

azeey (approved):

azeey:
Not sure why the test names have `<ini>` in them in https://build.osrfoundation.org/view/gz-ionic/job/gz_transport-ci_asan-main-jammy-amd64/32/, but this LGTM!
image

j-rivero:
> Not sure why the test names have `<ini>` in them in https://build.osrfoundation.org/view/gz-ionic/job/gz_transport-ci_asan-main-jammy-amd64/32/, but this LGTM!

Looking into the XML attributes I can not see anything obvious that can trigger this. I agree that it is not a blocker

PR #1134, opened 2024-05-06
Use noble on Ionic and __upcoming__ -ci- instead of Jammy

j-rivero:
Move -ci- jobs to run on noble from jammy for the Ionic releaes.


j-rivero:
superseeded by #1138

PR #1136, opened 2024-05-08
Use platform in docker call for amrhf builds on Noble

j-rivero:
Follow up of #1135. Needs platform command for ubuntu different


j-rivero:
Tested https://build.osrfoundation.org/job/gz-cmake3-debbuilder/1283/

PR #1140, opened 2024-05-16
Fix check for pub/sub in unofficial wrappers

j-rivero:
Mainly fix the comparison that was adding a double negative with `!` and `grep -v` . I added a gz topic info command that should generate time enough to the pub/sub system to work. 

Testing it here [Build Status](https://build.osrfoundation.org/job/ros_gzharmonic_bridge-install-pkg_humble-ci-jammy-amd64/66/)


Crola1702 (approved):

j-rivero:
That did not work. I've rework the timing a bit on [6aa82ce](https://github.com/gazebo-tooling/release-tools/pull/1140/commits/6aa82ceeb940483bd4cf8913cfad11cd05394f25). 

Tested [Build Status](https://build.osrfoundation.org/job/ros_gzharmonic_bridge-install-pkg_humble-ci-jammy-amd64/67/) and in the next one.

PR #1146, opened 2024-05-22
revert fix for testing purposes

scpeters:
don't merge, just testing that CI fails with the fix reverted


scpeters:
fails as expected

PR #1149, opened 2024-05-31
Fix null check in CI

j-rivero:
The null check in dsl_checks looks for the 'null' string in all the XML config generated. There are some jobs generated that include Groovy code that actually use null values. The PR modifies it to restrict the check to be the value under XML tags.


j-rivero:
Merging to unblock new PRs

PR #1148, opened 2024-05-31
Include a script to obtain gazebo releases for a given library

j-rivero:
A helper script to obtain the Gazebo collection/releases given a library and a major version. It uses the gz-collections.yaml as the database to search for the metadata.

PR #1150, opened 2024-06-03
Add support to report issue to ROS gz_*_vendor repositories in release.py

j-rivero:
Wait to process #1148 first

This PR adds support in `release.py` to automatically create issues in the gz_*_vendor repositories when a new release for any gazebo libraries is being created:

  * To detect if a vendor repository is affected by a the release in process, the release.py uses the script `jenkins-scripts/dsl/tools/get_collections_from_package_and_version.py` in #1148 that is able to return Gazebo releases given a library a version. To known the vendor repository from a Gazebo releases the script uses a simple approach by now defining a map of gazebo-release <-> ros-distro `ROS_VENDOR = {'harmonic': 'rolling'}` that assumes also that vendor repository branches are named in the same way than the ROS release names.
  * To create the issues the release.py code relies on calling the `gh` tool https://cli.github.com/ . If the command is not present in the system, the release.py script will ask the user to create the issue manually.
 
The PR also implements an output parsing for detecting if the ROS vendor commands are being processed correctly for a variety of options.


release.py
```diff
@@ -515,6 +517,64 @@ def display_help_job_chain_for_source_calls(args):
...
+           ' * Sync to the new version in CMakelists.txt \n'\
+           ' * Bump the patch version in package.xml \n'\
+           ' * Run the release process for this ROS package'
+    _out, _err = create_issue_in_repo(gz_vendor_repo, title, body)
+    if _err:
+        print('  !! An error happened running the "gh issue" cmd. Do not run this script again.\n'
+              '     Please create the issue manually. The error reported was:\n'
+              f'     {_err}')
+        sys.exit(1)
+    return _out
```
azeey, line 575: Is it possible to create a pull request instead? All we have to do is run https://github.com/gazebo-tooling/gz_vendor/blob/main/create_gz_vendor_pkg/create_vendor_package.py and give it the up-to-date package.xml from the upstream Gazebo library.

j-rivero:
> Is it possible to create a pull request instead? All we have to do is run https://github.com/gazebo-tooling/gz_vendor/blob/main/create_gz_vendor_pkg/create_vendor_package.py and give it the up-to-date package.xml from the upstream Gazebo library.

Moving to a new PR for the creation of automatically PRs for updating versions.

PR #1154, opened 2024-07-26
Missing Fortress/Garden jammy packaging in gz-collections.yaml

j-rivero:
@sauk2 discovered that we miss the Jammy packaging options for Garden and Fortress.


j-rivero:
Changes looks good now.

PR #1159, opened 2024-08-05
Add Noble packaging for Harmonic in gz-collections

j-rivero:
Missing Noble packaging configuration available for Harmonic


j-rivero:
There was a bug generating jammy versions instead of noble when pgk-config == noble was used. It is resolved and it has changed all the Ionic -install- builds

Issue #1092, opened 2023-12-05
Release Ionic


azeey:
## Timeline

Current timeline

* ~Feature freeze: 2024/07/26 (Friday)~
* ~Code freeze: 2024/08/28 (Wednesday)~
* ~Tutorial party: 2024/08/29 (Thursday) ~ 2024/09/13 (Friday)~
* ~Internal QA: 2024/09/16 (Monday) - 2024/09/20 (Friday)~
* ~Community meeting: 2024/09/25 (Wednesday)~
* Announcement: 2024/09/30 (Monday)

## During development

- [x] Bump major version of libraries early in the development cycle (use [bump_libraries](bump_libraries.md) template).
- [x] Work with artist on logo.
  - Artwork announced on https://community.gazebosim.org/t/gazebo-ionic-tutorial-party-results-artwork-swag-sale-and-more/3060/1

## Feature freeze

- [x] Merge each library forward from previous versions
- [x] Create stable branches off of `main`
  - [x] gz-cmake:
    - [x] Create a changelog PR ([example](https://github.com/gazebosim/gz-cmake/pull/444)): https://github.com/gazebosim/gz-cmake/pull/444
    - [x] `New branch` button on branches page ([example](https://github.com/gazebosim/sdformat/branches)). Make sure to set the `Source` to `main`.
    - [x] Enable CI for stable branches : https://github.com/gazebo-tooling/release-tools/pull/1164
    - [x] Update gazebodistro: https://github.com/gazebo-tooling/gazebodistro/pull/165
    - [x] Update homebrew: https://github.com/osrf/homebrew-simulation/pull/2730
    - [x] Run the Jenkins [DSL generation job](https://build.osrfoundation.org/job/_dsl_generate_all_dsls/)
    - [x] Create a small PR targetting the stable branch. This could be used to update build status badges ([example](https://github.com/gazebosim/gz-cmake/pull/445)). Don't skip this; it is important for the next step: https://github.com/gazebosim/gz-cmake/pull/445
    - [x] Update branch protections. Remember to include the abi-checker.
    - [x] Make a pre-release. Skip merging the homebrew PR. We want to keep formulas building from source until code freeze to match Ubuntu nightlies and windows.
   - [x] gz-utils:
      - [x] Create a changelog PR ([example](https://github.com/gazebosim/gz-cmake/pull/444)): https://github.com/gazebosim/gz-utils/pull/136
[...]

PR #1164, opened 2024-08-20
Use stable branch for gz-cmake4

azeey:
Part of https://github.com/gazebo-tooling/release-tools/issues/1092

Needed to make a pre-release of gz-cmake4.

### DIFF
--- a/jenkins-scripts/dsl/gz-collections.yaml
+++ b/jenkins-scripts/dsl/gz-collections.yaml
@@ -359,7 +359,7 @@ collections:
       - name: gz-cmake
         major_version: 4
         repo:
-          current_branch: main
+          current_branch: gz-cmake4
       - name: gz-tools
         major_version: 2
         repo:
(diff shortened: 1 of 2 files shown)


jenkins-scripts/dsl/gz-collections.yaml
```diff
@@ -445,6 +445,10 @@ collections:
         major_version: 3
         repo:
           current_branch: main
+      - name: gz-cmake
+        major_version: 4
```
scpeters, line 449: should this be 5?

azeey: We don't have a branch with that yet. If we want to this to be 5, I think `main` will need to be bumped. I was hoping not to go through version bumps because it's a lot of effort.

scpeters: this is the `__upcoming__` section, which currently only has `gz-tools3`, so I suggested bumping the major version based on pattern matching, not based on how I think `__upcoming__` should be used. This is fine as is

scpeters (approved):

PR #1173, opened 2024-08-26
Use stable branch gz-transport14

caguero:
Part of https://github.com/gazebo-tooling/release-tools/issues/1092

Needed to make a pre-release of gz-transport14.

### DIFF
--- a/jenkins-scripts/dsl/gz-collections.yaml
+++ b/jenkins-scripts/dsl/gz-collections.yaml
@@ -399,7 +399,7 @@ collections:
       - name: gz-transport
         major_version: 14
         repo:
-          current_branch: main
+          current_branch: gz-transport14
       - name: gz-gui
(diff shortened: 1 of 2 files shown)


jenkins-scripts/dsl/gz-collections.yaml
```diff
@@ -469,6 +469,10 @@ collections:
         major_version: 9
         repo:
           current_branch: main
+      - name: gz-transport
+        major_version: 14
+        repo:
+          current_branch: gz-transport14
```
iche033, line 475: I think this should be `main`?
```suggestion
          current_branch: main
```

caguero: Thanks, good catch!

iche033 (approved):

PR #1175, opened 2024-08-27
source_changelog.bash: print baseline version

scpeters:
We'd like to print the baseline version for the changelog of a given release, so print a formatted line with this info in our changelog script.


scpeters:
tested with https://github.com/gazebosim/sdformat/pull/1473

azeey (approved):

PR #1190, opened 2024-10-15
Replace /usr/local with HOMEBREW_PREFIX part 1

scpeters:
One of our brew build scripts has a `HOMEBREW_PREFIX` variable to represent the location where brew is installed (either `/usr/local` for Intel macOS machines or `/opt/homebrew` for arm macOS machines), but it also has many paths hard-coded to /usr/local. In preparation for supporting arm-based CI machines, use the `HOMEBREW_PREFIX` variable more widely.

### DIFF
--- a/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
+++ b/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash
@@ -24,7 +24,7 @@ export HOMEBREW_PREFIX=/usr/local
 export HOMEBREW_CELLAR=${HOMEBREW_PREFIX}/Cellar
 export PATH=${HOMEBREW_PREFIX}/bin:${HOMEBREW_PREFIX}/sbin:$PATH
 
-export PYTHONPATH=$PYTHONPATH:/usr/local/lib/python
+export PYTHONPATH=$PYTHONPATH:${HOMEBREW_PREFIX}/lib/python
 
 # make verbose mode?
 MAKE_VERBOSE_STR=""
@@ -90,7 +90,7 @@ if [[ -n "${PIP_PACKAGES_NEEDED}" ]]; then
   brew install python3
   PIP=pip3
   if ! which ${PIP}; then
-    PIP=/usr/local/opt/python/bin/pip3
+    PIP=${HOMEBREW_PREFIX}/opt/python/bin/pip3
   fi
   # TODO use a python3 venv instead.
   ${PIP} install --break-system-packages ${PIP_PACKAGES_NEEDED}
@@ -98,7 +98,7 @@ fi
 
 if [[ -z "${DISABLE_CCACHE}" ]]; then
   brew install ccache
-  export PATH=/usr/local/opt/ccache/libexec:$PATH
+  export PATH=${HOMEBREW_PREFIX}/opt/ccache/libexec:$PATH
 fi
 echo '# END SECTION'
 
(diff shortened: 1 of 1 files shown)


claraberendsen (commented):
Thanks @scpeters for setting this up! A quick question related to the name of the bash script ... since we are generalizing it, would it make sense to update the name to be `project-default-devel-homebrew.bash`?  
I'm not sure if there are more  particularities in the script that would make keeping the `amd64` naming still worth it.

scpeters:
> Thanks @scpeters for setting this up! A quick question related to the name of the bash script ... since we are generalizing it, would it make sense to update the name to be `project-default-devel-homebrew.bash`? I'm not sure if there are more particularities in the script that would make keeping the `amd64` naming still worth it.

I'm fine with changing the name of the script, though it's not quite generalized yet.

scpeters:
> > Thanks @scpeters for setting this up! A quick question related to the name of the bash script ... since we are generalizing it, would it make sense to update the name to be `project-default-devel-homebrew.bash`? I'm not sure if there are more particularities in the script that would make keeping the `amd64` naming still worth it.
> 
> I'm fine with changing the name of the script, though it's not quite generalized yet.

I'd also prefer to change it in a separate pull request to keep the changes from getting mixed together

j-rivero:
> I'd also prefer to change it in a separate pull request to keep the changes from getting mixed together

+1

j-rivero:
we have a few more places that use usr/local/bin for populating the BREW_BINARY_DIR variable:
```bash
release-tools/jenkins-scripts on  scpeters/brew_prefix [?] ❯ grep -r local/bin  lib/
lib/_homebrew_base_setup.bash:BREW_BINARY_DIR=/usr/local/bin
lib/homebrew_bottle_creation.bash:export PATH="/usr/local/bin:$PATH"
lib/_homebrew_cleanup.bash:BREW_BINARY_DIR=/usr/local/bin
```

Do we need to change these too?

scpeters:
> we have a few more places that use usr/local/bin for populating the BREW_BINARY_DIR variable:
> 
> ```shell
> release-tools/jenkins-scripts on  scpeters/brew_prefix [?] ❯ grep -r local/bin  lib/
> lib/_homebrew_base_setup.bash:BREW_BINARY_DIR=/usr/local/bin
> lib/homebrew_bottle_creation.bash:export PATH="/usr/local/bin:$PATH"
> lib/_homebrew_cleanup.bash:BREW_BINARY_DIR=/usr/local/bin
> ```
> 
> Do we need to change these too?

The challenge that I didn't try to address here is how to get the `brew` command in your `PATH`, which is what most of these instances are doing. When you install `brew` on Intel to `/usr/local`, it should already be in your path, so you can just execute `brew --prefix`, but on arm64 it installs to `/opt/homebrew` and recommends in the [console output after installation](https://github.com/Homebrew/install/blob/cc1a94045eeacbb80bfc37250f1659f542bc5c83/install.sh#L1047-L1063) to `eval` the [brew shellcheck](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/shellenv.rb) command in a terminal or in your shell rc file to add `brew` to your `PATH`.

I feel like that is something that should be preconfigured on the build machine but am not sure where we currently configure that. Once we've addressed that, we can clean this up further. For this PR, I was just cleaning up a file that already defines `HOMEBREW_PREFIX` at the top of the file but wasn't using it throughout. So this is an incremental step.

j-rivero:
> I feel like that is something that should be preconfigured on the build machine but am not sure where we currently configure that.

There are environments variable that we can configure per-node when we provision them. We probably could use that.

> For this PR, I was just cleaning up a file that already defines HOMEBREW_PREFIX at the top of the file but wasn't using it throughout. So this is an incremental step.

Understood, thanks for clarifying.

j-rivero (approved):

Issue #373, opened 2021-01-12
Migrate Windows CI from vcpkg to conda-forge


j-rivero:
We are using vcpkg to implement our CI on Windows, particularly to get external dependencies and inject them using `-DCMAKE_TOOLCHAIN_FILE` into `colcon` to build all ignitions from source.

Our documentation is recommending to use conda-forge for having binaries installed on Windows and would be good to move our CI infrastructure to follow the tools we recommend for end-user experience. One option could be to replicate the vcpkg setup and install the dependencies and use colcon to build all libraries together from source. A second step could involve to have conda-forge packages for every ignition library and only build from source the library under testing (like we do in Ubuntu with apt).

PR #1192, opened 2024-10-31
Initial conda configurations import (pixi testing on github actions)

j-rivero:
The PR adds the initial conda/pixi configurations together with a README explaining the directory structure and the files contained.


conda/envs/legacy/pixi.toml
```diff
@@ -0,0 +1,64 @@
...
+gflags = "*"
+glib = "*"
+gts = "*"
+jsoncpp = "*"
+libccd-double = "*"
+libprotobuf = "*"
+libsqlite = "*"
+libzip = "*"
+ogre = "1.10.12.1.*"
+ogre-next = "2.3.3.*"
```
azeey, line 31: The current version in vcpkg is 2.2.x. Using 2.3.3 here will break Fortress builds. Should we not have `envs/fortress`, `envs/harmonic`,  `envs/ionic`, and `envs/jetty`?

conda/envs/legacy/gz-environment.yaml
azeey, line 1: Is this file necessary if we have the `pixi.toml` file?

conda/envs/legacy/pixi.toml
j-rivero: This first step tries to reproduce the legacy environment created by vcpkg and afterwards to create new environments for each (new?) distribution. The reason is not to alter too much the current status of the existing Windows CI builds with respect about how they are on vcpkg. I changed the version in [211a78a](https://github.com/gazebo-tooling/release-tools/pull/1192/commits/211a78aaec76c9e1aa3bd9f4b094abd539443d63)

conda/envs/legacy/gz-environment.yaml
j-rivero: It is not in use by us. It was the first file designed to work with the other mamba/micromamba/conda package managers. We could just remove it if we are not going to use it.

azeey: I'd rather avoid duplicating the information unless it's truly necessary.

conda/envs/legacy/pixi.toml
azeey: Okay, that makes sense. For the follow-up work, I would really like gz-rendering CI for Harmonic and Ionic to run with Ogre 2.3.3 and resolve https://github.com/gazebosim/gz-rendering/issues/745 so it would be great if we created new environments for existing distributions as well as new ones.

j-rivero: Understood. So custom environments start with Harmonic and Ionic, maybe the same than legacy but using ogre-2.3.3 ?

azeey: Yes, I think that's right. The only other version that might be different is CMake, but I would try with the latest first and see if we have any problems.

conda/envs/legacy/pixi.toml
```diff
@@ -0,0 +1,60 @@
...
+libwebsockets = "*"
+mesa-libegl-devel-cos7-x86_64 = "*"
+mesa-libgl-devel-cos7-x86_64 = "*"
+vulkan-headers = "*"
+
+[target.linux-aarch64.dependencies]
+libglu = "*"
+libglvnd-devel-cos7-aarch64 = "*"
+mesa-libegl-devel-cos7-aarch64 = "*"
+mesa-libgl-devel-cos7-aarch64 = "*"
```
azeey, line 54: If this is mainly for windows, do we need this section?

j-rivero: I don't think so. It is useful for linux users and as a long-term goal we probably want to have the platforms unified in a single source of truth so I added it here even if we are not testing/supporting it as the README mention. Hopefully it can receive patches or updates from users.

azeey (approved):

azeey:
xref https://github.com/gazebo-tooling/release-tools/issues/373

PR #1199, opened 2024-11-06
Add jetty collection to gz-collections.yaml

Crola1702:
As the title says


j-rivero:
If I'm not wrong the appointed branches for Jetty are already setup in https://github.com/gazebo-tooling/gazebodistro/pull/181 and should be the same here. So no `main` branches but probably starting with Ionic branches.

Crola1702:
> So no main branches but probably starting with Ionic branches.

Done. Can you check if everything is correct? @j-rivero

j-rivero (changes requested):
I think that major_version are not in sync with the package names. I.e: gz-cmake4 branch for gz-cmake5 package. So we need to update the major_versions too.

jenkins-scripts/dsl/gz-collections.yaml
```diff
@@ -441,67 +441,67 @@ collections:
       - name: gz-cmake
         major_version: 5
         repo:
-          current_branch: gz-cmake4
+          current_branch: gz-cmake5
```
azeey, line 444: We should do the opposite here. Currently, Jetty should be the exact copy of Ionic.

j-rivero (approved):

PR #1201, opened 2024-11-08
Change auth method in releasepy: remove shared token and use Jenkins API tokens

j-rivero:
Making @nuclearsandwich happy. Remove the token used in DSL.

* Change to remove the existing token [19f9405](https://github.com/gazebo-tooling/release-tools/pull/1201/commits/19f9405526ecb6b5ea3c9efb155c434f82c2c30f)
* Change the authentication model to use per-user Jekins API tokens: [41d75d8](https://github.com/gazebo-tooling/release-tools/pull/1201/commits/41d75d881e0268fc53b4cdf586ccfb38051020aa)

Developers documentation update: https://github.com/gazebosim/docs/pull/538

This change will break existing nightly scheduler calls. A subsequent PR should take care of it.

PR #353, opened 2020-12-04
Use credential bindings for osrfbuild github token

j-rivero:
The PR removes the action of taking the `GITHUB_TOKEN` from the filesystem and implements the credentials plugin from Jenkins on it. There is also a testing job generated that check permission and push capabilities running directly on osX nodes: [Build Status](https://build.osrfoundation.org/job/_test_job_osrfbuild-credentials-token_from_dsl/7/)

Linked to: https://github.com/osrf/chef-osrf/pull/55

PR #1206, opened 2024-11-14
Change nightly call to support new API tokens

j-rivero:
Build on top of #1201.

**1. Adapt the release.py to not use the jenkins.ini file**

The PR prepares the new auth method to work with nightly generation. To call release.py with osrfbuild bot account this PR implements the `--auth` parameter that supports `user:token` to be passes as argument to authenticate in our Jenkins. The token can potentially be a Jenkins API token or a GitHub Token. d7dcb2d831fcd610e1d1f76b511bf59ad8bed014

This has the side effect on the test code that we no longer need the `_RELEASEPY_TEST_CREDENTIALS`, removed in af2ecae98bfa787a0c02f202e6a2bcea8ca6a676.

**2. Change release-tools DSL to modify the nightly_scheduler code**

Remove existing shared token injection and use credentials plugin in Jenkins to pass GITHUB_TOKEN. Use that with the `--auth` mechanism implemented in 1. [85ea888](https://github.com/gazebo-tooling/release-tools/pull/1206/commits/85ea888c5eb0332fbdcba1f1b56f89df03b6eac2)

**3. Test DSL to check that token is able to login in Jenkins and use the osrfbuild fork**

Hijack part of #353 and add a test to check that that the token is able to login into Jenkins.  [07a29cb](https://github.com/gazebo-tooling/release-tools/pull/1206/commits/07a29cb926cb7e32beea5b9fd2e2b81af2f8921c). The new token should have the fine-grained permissions on osrfbuild/homebrew-simulation fork for "Contents".

:warning:  Existing problem

The existing `GITHUB_TOKEN` for the osrfbuild user in Jenkins does not seem to work to authenticate into Jenkins. This links with the old PR #353 that probably needs to be resurrected first.

**Needs to merge solution to this in https://github.com/gazebo-tooling/release-tools/pull/1212 before**


j-rivero:
> The other main question is: do we prefer to use an existing token from GitHub to authenticate into Jenkins with the osrfbuild or a Jenkins API Token. 

After speaking with @nuclearsandwich, we agreed on using a manually generated Jenkins API Token. It is much better scoped for the task. I've added it and test in https://github.com/gazebo-tooling/release-tools/pull/1212

j-rivero:
Cherry picking changes to #1201 manually. Closing this.

PR #1207, opened 2024-11-14
Add testing builds for Conda on windows for fortress using DSL

j-rivero:
The changes should generate parallel CI works for Fortress on Windows named `-cwin` instead of just `-win` that should allow us to test and compare results. Note that the builds have custom config to run only in `test-instance` label and set the USE_PIXI paramter.

In draft since we probably don't want to support officially both builds at the same time but replacing ones with the others. The PR also allows to check what is going to be generated by the DSL.


azeey:
xref https://github.com/gazebo-tooling/release-tools/issues/373

j-rivero:
Not useful anymore. Closing.

PR #1208, opened 2024-11-14
Conda legacy env: use vs2019 package and fixed versions

j-rivero:
To reduce the number of changes brought by `pixi update` on each iteration, I'm setting some of the versions in pixi.toml for part of the relevant software.

While testing this, I noticed that the we can declare which msvc compiler to use via toml configuration (that previously was included as a dependency by one of versions used by the explicit set of dependencies in this toml file). I'm declaring it explicitly here until we decide in we want to use compilers this way or use the system installed ones.

### DIFF
--- a/conda/envs/legacy/pixi.toml
+++ b/conda/envs/legacy/pixi.toml
@@ -8,14 +8,14 @@ platforms = ["linux-64", "linux-aarch64", "win-64"]
 
 [dependencies]
 assimp = "*"
-bullet-cpp = "*"
+bullet-cpp = "3.25"  # compatible with dart
 cppzmq = "*"
 curl = "*"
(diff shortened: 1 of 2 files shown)


conda/envs/legacy/pixi.toml
```diff
@@ -26,18 +26,19 @@ libzip = "*"
...
 qwt = "*"
 ruby = "3.2.*"
 tinyxml2 = "*"
 yaml = "*"
-zeromq = "*"
+zeromq = "4.3.4"
 
 [target.win-64.dependencies]
 dlfcn-win32 = "*"
+vs2019_win-64 = "*"
```
j-rivero, line 41: This is not the real compiler but the runtime libraries, we probably can leave it out.

PR #1209, opened 2024-11-14
Windows colcon: relocate Fortress workaround for ignition names

j-rivero:
Preparing to execute conda enviroments, the existing colcon script does use of colcon as if it was installed in the system. While this is true in the current provisioning, for conda we use colcon declared in the toml file what also facilitates the use of it locally.

The PR relocates the code that helps with the existing `ignition` names (Gz Fortress) into a section after the dependencies/environment files are installed. 

Testing it here [Build Status](https://build.osrfoundation.org/job/gz_sim-pr-win/1042/)


j-rivero:
Fixed and re-tested [Build Status](https://build.osrfoundation.org/job/gz_sim-pr-win/1044/)

j-rivero:
Merging to unblock other Conda updates

PR #1212, opened 2024-11-20
Add OSRFcredentials class and test for DSL

j-rivero:
This PR replaces the `test.dsl` in #1206 and checks the two main token that are going to be used for the `osrfbuild` user as discussed in https://github.com/gazebo-tooling/release-tools/pull/1206#issuecomment-2483688025:

 * OSRFBUILD_GITHUB_TOKEN: for git operations and pull requests
 * OSRFBUILD_JENKINS_TOKEN: for authenticate into the server
 
The credential class also has an `allowOsrfbuildToRunTheBuild` that can restrict which projects can be build by the `osrfbuild` user when making remote calls.

Operations needs to be done in production:
 * Generate a new `OSRFBUILD_JENKINS_TOKEN` for the `osrfbuild` user with the id of `
 * Change the type of the `OSRFBUILD_GITHUB_TOKEN` or generate a new one and set it up as a `usernamePassword` with the ID `github-osrfbuild-credentials`

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFCredentials.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFCredentials.groovy
@@ -15,7 +15,7 @@ class OSRFCredentials
             if (credential_keyword == 'OSRFBUILD_GITHUB_TOKEN') {
                 usernamePassword('OSRFBUILD_USER',
                                  'OSRFBUILD_TOKEN',
-                                 'github-osrfbuild-credentials')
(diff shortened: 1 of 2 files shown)


j-rivero:
Working for both Jenkins API and Github token in https://build.osrfoundation.org/job/_test_job_osrfbuild-credentials-token_from_dsl/14/badge/

j-rivero:
Going to merge to unblock other credential migration.

Issue #1223, opened 2024-12-13
EOL collection: Citadel


azeey:
- [x] Write a post to https://community.gazebosim.org/ announcing the community it has last X weeks for finishing their PRs.
    * https://community.gazebosim.org/t/gazebo-classic-and-citadel-end-of-life/3242/4
- [x] Check library versions that can be EOL’d on the [versions table](https://github.com/gazebosim/docs/blob/master/tools/versions.md)
    - gz-sim3
    - gz-gui3
    - gz-launch2
    - gz-physics2
    - gz-rendering3
    - gz-sensors3
- [x] Make releases with outstanding changes in each library about to be EOL'ed
    - https://github.com/gazebosim/gz-sim/pull/2700
    - https://github.com/gazebosim/gz-gui/pull/649
    - ~gz-launch2~
    - ~gz-physics2~
    - ~gz-rendering3~
    - ~gz-sensors3~
- [x] Archive release repos in [gazebo-release](https://github.com/gazebo-release) for all libraries after the last release has been made
  - [x] https://github.com/gazebo-release/gz-physics2-release
  - [x] https://github.com/gazebo-release/gz-rendering3-release
  - [x] https://github.com/gazebo-release/gz-sensors3-release
  - [x] https://github.com/gazebo-release/gz-gui3-release
  - [x] https://github.com/gazebo-release/gz-sim3-release
  - [x] https://github.com/gazebo-release/gz-launch2-release
- [x] Update `release-tools` after all releases have been made ([example](https://github.com/gazebo-tooling/release-tools/pull/369))
    * https://github.com/gazebo-tooling/release-tools/pull/1228
- [x] Deprecate homebrew bottles if they aren't deprecated yet (don't disable them) ([example](https://github.com/osrf/homebrew-simulation/pull/1785))
    * https://github.com/osrf/homebrew-simulation/pull/1789 (Done previously, thanks @scpeters!)
- [x] Remove badges from `homebrew-simulation` README ([example](https://github.com/osrf/homebrew-simulation/pull/1772))
    * https://github.com/osrf/homebrew-simulation/pull/2919
- [x] Remove from buildfarmer ([example](https://github.com/osrf/homebrew-simulation/pull/1785))
    * Done automatically
- [x] Remove from `pr-collection-labeler` ([example](https://github.com/gazebo-tooling/pr-collection-labeler/pull/13))
    * https://github.com/gazebo-tooling/pr-collection-labeler/pull/52
[...]

PR #1228, opened 2024-12-17
🏁 Citadel EOL

azeey:
Part of https://github.com/gazebo-tooling/release-tools/issues/1223

**DO NOT MERGE UNTIL Jan 1, 2025**


caguero (commented):
Remove this example line? https://github.com/gazebo-tooling/release-tools/blob/master/bloom/ros_gz/README.md?plain=1#L18

j-rivero:
> > Remove this example line? https://github.com/gazebo-tooling/release-tools/blob/master/bloom/ros_gz/README.md?plain=1#L18
> 
> I think this can be left as is, since it's just an example of officially supported ROS/Gazebo pairings. Even though Citadel is EOL, the fact that it was paired with ROS 2 Foxy has not changed.

Not having a strong position here but I usually clean up most references to EOL distributions just to reduce the noise for the people reading the options. I'm fine with it to stay.

>  Add the Foxy - Citadel combination to the List of EOL releases section? https://github.com/gazebo-tooling/release-tools/blob/master/bloom/ros_gz/README.md?plain=1#L37
>  This list is for unofficial Gazebo/ROS pairings, e.g. Harmonic/Humble while the official one is Fortress/Humble. Since Citadel never had an unofficial pairing, I don't think there's anything to do here.

:+1: 

>     Remove the citadel block from _dashboard_lib.sh? https://github.com/gazebo-tooling/release-tools/blob/master/terminal-dashboard/_dashboard_lib.sh#L9-L26
>    Remove the citadel block from table.bash? https://github.com/gazebo-tooling/release-tools/blob/master/terminal-dashboard/table.bash
> Is there a reason to do so? Even if it's EOL, I think it's nice to have the ability to generate the citadel dashboard. The table command takes the distro as an argument, so users are not forced to generate Citadel's table if they don't want to.

Same than in first point, not an strong argument but we probably remove previous versions from the table to reduce the number of lines of code that the dashboard is hosting. I'm also fine with it to stay for some months just in case someone needs it a some point.

j-rivero (approved):

PR #1229, opened 2025-01-07
Implement GPU checking on Windows nodes

j-rivero:
Add support for the `GPU_SUPPORT_NEEDED` flag on Windows that is used to check a right configuration of the Nvidia GPU card. This can safe time debugging when the problem is in the Jenkins node configuration. Replaces existing call on dxdiag unconditionally.

Waiting to be tested:
[Build Status](https://build.osrfoundation.org/job/gz_rendering-pr-win/287/)


j-rivero:
In action here https://build.osrfoundation.org/job/gz_rendering-pr-win/287/consoleFull#console-section-0

PR #1230, opened 2025-01-10
Add cli11 to VCPKG_DEPENDENCIES_LEGACY and Conda legacy env

scpeters:
Needed by gazebosim/gz-utils#167.


j-rivero:
Version in our vcpkg snapshot is "2.1.2" https://github.com/microsoft/vcpkg/blob/2022.02.23/ports/cli11/vcpkg.json. Is it good enough? We have the same version available in Conda.

scpeters:
> Version in our vcpkg snapshot is "2.1.2" https://github.com/microsoft/vcpkg/blob/2022.02.23/ports/cli11/vcpkg.json. Is it good enough? We have the same version available in Conda.

the [current vendored version is 1.9.1](https://github.com/gazebosim/gz-utils/blob/gz-utils3/cli/README.md?plain=1#L5), so yes I think 2.1.2 is good enough

scpeters:
still waiting for an approval for this

iche033 (approved):

PR #1248, opened 2025-01-27
Only check Conda environments on USE_PIXI

j-rivero:
Should make the existing jobs not to fail

//cc @Crola1702 sorry for that


j-rivero:
Tested manually here [Build Status](https://build.osrfoundation.org/job/gz_rendering-main-win/122/)

j-rivero:
Going to merge and re-run dsl since existing jobs are broken

PR #1255, opened 2025-01-30
Add script that helps with preparing a library for release

azeey:
- Bumps the version CMakeLists
- Bumps the version package.xml if present
- Updates the changelog file

### DIFF
--- a/source-repo-scripts/prepare_for_release.py
+++ b/source-repo-scripts/prepare_for_release.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2024 Open Source Robotics Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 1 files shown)


source-repo-scripts/prepare_for_release.py
```diff
@@ -0,0 +1,232 @@
...
+    else:
+        repo = m.group(0)
+        return repo, REPO_NAMES[repo]
+
+
+def calculate_new_version(bump: str, previous_version: str):
+
+    new_version = [int(v) for v in previous_version.split(".")]
+    if bump == "major":
+        new_version[0] += 1
```
caguero, line 90: I think that when bumping the major version we also need to set minor and patch to `0`. Same when bumping the minor version, we need to set patch to `0`.

caguero (changes requested):

PR #1262, opened 2025-02-14
releasepy: push only the tag of the release

j-rivero:
release.py was using git push with the `--tags` option, what caused to push all the tags to the remote. This was causing troubles to @caguero with old tags that were not matching the remote tags. We can limit the action to just the push of the generated tag, this is what this PR is doing.

Needs manual testing before merging.

### DIFF
--- a/release.py
+++ b/release.py
@@ -510,7 +510,7 @@ def tag_repo(args):
         # tilde is not a valid character in git
         tag = '%s_%s' % (args.package_alias, args.version.replace('~', '-'))
         check_call(['git', 'tag', '-f', tag])
-        check_call(['git', 'push', '--tags'])
+        check_call(['git', 'push', 'origin', 'tag', tag])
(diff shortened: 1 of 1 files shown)


release.py
```diff
@@ -516,7 +516,7 @@ def tag_repo(args):
         # tilde is not a valid character in git
         tag = '%s_%s' % (args.package_alias, args.version.replace('~', '-'))
         check_call(['git', 'tag', '-f', tag])
-        check_call(['git', 'push', '--tags'])
+        check_call(['git', 'push', 'origin', 'tag', tag])
```
azeey, line 519: This assumes the remote is called `origin`. I used to name mine `upstream`, but I have sinced changed since a lot of our tools assume `origin`, but I'm not sure if it's a necessary assumption to make. Would this work without specifying the remote?

PR #1264, opened 2025-02-18
Avoid duplicating packages in colcon Windows CI

j-rivero:
I've found some weird reports where the pull request seemed to be used an old version of the code. I found that the clean up of the colcon workspace on Windows was not done in the right way. 

Tested [Build Status](https://build.osrfoundation.org/job/gz_common-pr-clowin/4/)


j-rivero:
Tested in gz-cmake3. Works fine.

PR #1263, opened 2025-02-17
Cleanup: remove unused sdformat/rndf scripts

j-rivero:
Removing obsolete files. Can not find references to them in the source code and there is a check in CI to look for existing scripts referenced in the XML configurations.

PR #1267, opened 2025-02-21
Fix CI on missing scripts check

j-rivero:
The CI can detect some occasions when the Jenkins scripts are calling a file that is not in the repository anymore. This should have happened in #1263 and although the script detected the problem it did not fail correctly.

The changes to correct this were in #1258 but I'm porting them here so it does not happen again while that PR is being merged. I expect the CI to fail, should be corrected by #1266


j-rivero:
The PR is correct, it detected the problem.

PR #1268, opened 2025-02-21
Check for existing files in scripts called inside docker/

j-rivero:
Add a CI step for checking that our Linux scripts are calling real existing files in this repository. This is a bit different of our existing DSL check that checks that XML Jenkins configuration are initially calling an existing files but do not follow the chain of subcalls that a script can execute.


j-rivero:
The fail is correct. It needs https://github.com/gazebo-tooling/release-tools/pull/1269

PR #1270, opened 2025-02-24
Make Windows colcon scripts to support EXTRA_EXIT_PARAM 

j-rivero:
Preparing the terrain to land the local builder tool, the hard exit that we produce in Jenkins will kill the whole Windows terminal in a local system. The variable permits to inject a `\b` parameter that will not close the whole session but just fail in an standard way.


j-rivero:
Tested [Build Status](https://build.osrfoundation.org/view/Conda/job/gz_common-pr-clowin/5/)

PR #1272, opened 2025-02-27
Restrict nightly node labeler to linux (docker) only

j-rivero:
The nightly node selector was using plain `large-memory` label as criteria. After the inclusion of the Windows nodes with GPU and large-memory labels the nightly node was broken selecting Windows machines.

The PR adds the docker (linux only) tag to the expression. That required to modify the searches a bit to support label expressions and not single label matches.


scpeters:
does the DSL CI test cover the code that is changed here?

j-rivero:
> does the DSL CI test cover the code that is changed here?

It does not. It only works on generated XML Jenkins jobs https://github.com/gazebo-tooling/release-tools/blob/master/jenkins-scripts/dsl/dsl_checks.bash#L105-L118

Crola1702 (approved):

PR #1278, opened 2025-03-11
Change deprecated getEnvVars Groovy method in NO_MAILS

j-rivero:
reduce the log noise


j-rivero:
Testing it here [Build Status](https://build.osrfoundation.org/job/gz_cmake-ci-pr_any-noble-amd64/68/)

PR #1282, opened 2025-03-14
Fix pixi hooks run: do not use call with windows redirection

j-rivero:
To activate the pixi enviroment in our Jenkins the best way I found was to use the `shell-hook` command to create a .bat file that we can "call". For doing it the current code uses the `call %win_lib% :pixi_cmd ... > hooks.bat`. This is making that many of the scripts code ended up in the .bat file and it is generating random commands.

The PR changes the call to use directly the pixi executable. This generates a clear execution.


j-rivero (commented):
Tested here [Build Status](https://build.osrfoundation.org/job/gz_common-pr-clowin/9/)

PR #1285, opened 2025-03-17
Tools to generate buildfarm statistics

j-rivero:
Still in the form of python notebooks. Not to be merge but can be played with if someone is interested.

Download the data from: https://build.osrfoundation.org/job/_agent_usage_statistics/ws/agent_data.csv/*view*/ (need to be logged into Jenkins).

Probably the most useful data right now is the usage over the weekend since we don't have a full week logged yet from Monday to Friday in the log.
image

For the weekdays, still more data needs to be fetched but:
image

Need to group times by label to obtain an absolute measure.


j-rivero:
I was able to generate statistics by merging the times for the different labels:

image

j-rivero:
Context: nightly node creation starts at 09:00 server time (UTC) and the nightly and can run for four and a half hours until building the whole family (ends around 2:30). During that time the two Linux GPU nodes are busy full time. That is generating a 18.75% of Usage on each node linux-gpu each day. 

Range from 2025-03-12 18:55:02 ... 2025-03-31 23:55:01 . Using the filter out of status different of Busy/Idle in [6b9c3c6](https://github.com/gazebo-tooling/release-tools/pull/1285/commits/6b9c3c6eeb640abf0dbdba941ea145881ea6fca8)

image

And busy times by hour.

image

image


The busy times raise at the times of the nightlies, that is consistent with the expectation. I'm still a bit surprised about the high rates on weekends.

PR #1289, opened 2025-03-20
Retry jobs failed that have another brew update process

Crola1702:
Jobs that fail with error: `Error: Another ``brew update`` process is already running.` (e.g., https://build.osrfoundation.org/job/gz_sensors-ci-ign-sensors6-homebrew-amd64/123/console), can be retriggered safely

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFBase.groovy
@@ -38,9 +38,15 @@ class OSRFBase
           }
         }
 
-        // Create the naginator retry tag
+        // Create the naginator retry tags
         HelperRetryFailures.create(job, [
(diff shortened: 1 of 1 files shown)


j-rivero:
> Jobs that fail with error: Error: Another ``brew update`` process is already running. (e.g., https://build.osrfoundation.org/job/gz_sensors-ci-ign-sensors6-homebrew-amd64/123/console), can be retriggered safely

I see no problem on retrying the build once with the expectation that another node takes the new build and finished the process. It can improve some responses. 

The root problem however is not solve and still can cause more problems.  My experience is that most of the times we need to manually go to the machine to kill the zombie process. Would be great if we can disable the node and retrigger the build. The PR does part of this so I'm fine.

@scpeters what do you think?

scpeters (approved):
yeah, this is reasonable. I'm not sure why the issue happens

PR #1299, opened 2025-04-10
Use new pixi version 0.44.0

j-rivero:
There are problems with the pixi lock format when running modern pixi versions locally. The PR bumps the use of pixi to 0.44.0.


j-rivero:
Tested it [Build Status](https://build.osrfoundation.org/job/gz_utils-pr-clowin/3/)

Issue #1292, opened 2025-03-31
Bump in Jetty: gz-transport15, gz-sensors10


scpeters:
Libraries being bumped:

* gz-transport 15, gz-sensors 10: add Zenoh support
    - [x] Source code of all downstream libraries ([example](https://github.com/gazebosim/gz-transport/pull/149))
        - [x] https://github.com/gazebosim/gz-gui/pull/674
        - [x] https://github.com/gazebosim/gz-sensors/pull/515
        - [x] https://github.com/gazebosim/gz-sim/pull/2857
        - [x] https://github.com/gazebosim/gz-launch/pull/294
        - [x] https://github.com/gazebosim/gz-jetty/pull/6
    - [x] Release repositories of all downstream libraries ([example](https://github.com/gazebo-release/gz-fuel-tools7-release/pull/1))
        - [x] https://github.com/gazebo-release/gz-gui10-release/pull/4
        - [x] https://github.com/gazebo-release/gz-sensors10-release/pull/2
        - [x] https://github.com/gazebo-release/gz-sim10-release/pull/7
        - [x] https://github.com/gazebo-release/gz-launch9-release/pull/3
    - [x] Use `main` branch on `gazebodistro` ([example](https://github.com/gazebo-tooling/gazebodistro/pull/42))
        - [x] https://github.com/gazebo-tooling/gazebodistro/pull/186
        - [x] https://github.com/gazebo-tooling/gazebodistro/pull/187
        - [x] https://github.com/gazebo-tooling/gazebodistro/pull/188
        - [x] https://github.com/gazebo-tooling/gazebodistro/pull/189
        - [x] https://github.com/gazebo-tooling/gazebodistro/pull/190
    - [x] Build nightlies from the `main` branch ([example](https://github.com/gazebo-tooling/release-tools/pull/437))
        - [x] https://github.com/gazebo-tooling/release-tools/pull/1303
    - [x] homebrew-simulation: create formula and update dependencies ([example](https://github.com/osrf/homebrew-simulation/pull/14230))
        - [x] https://github.com/osrf/homebrew-simulation/pull/3008
        - [x] https://github.com/osrf/homebrew-simulation/pull/3009
        - [x] https://github.com/osrf/homebrew-simulation/pull/3015
        - [x] https://github.com/osrf/homebrew-simulation/pull/3016
        - [x] https://github.com/osrf/homebrew-simulation/pull/3017
[...]

PR #1304, opened 2025-04-15
Add example command to issue template

scpeters:
Example command used to open pull requests for #1292.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bump_libraries.md
+++ b/.github/ISSUE_TEMPLATE/bump_libraries.md
@@ -15,7 +15,7 @@ instead.
 
 This script can be used to do most of the work:
 
-https://github.com/gazebo-tooling/release-tools/blob/master/release-repo-scripts/bump_dependency.bash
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bump_libraries.md
```diff
@@ -17,6 +17,18 @@ This script can be used to do most of the work:
 
 https://github.com/gazebo-tooling/release-tools/blob/master/release-repo-scripts/bump_dependency.bash
```
iche033, line 18: I was checking what the args are and noticed that the path of this file incorrect: 

```suggestion
https://github.com/gazebo-tooling/release-tools/blob/master/source-repo-scripts/bump_dependency.bash
```

iche033 (approved):
minor comment on the path of the bash script but it's not part of the PR changes so we can do that as a separate PR if we want.

.github/ISSUE_TEMPLATE/bump_libraries.md
scpeters: wow, I don't know where that came from

PR #1306, opened 2025-04-17
Set -DCMAKE_POLICY_VERSION_MINIMUM in homebrew

Crola1702:
In gz_tools1 is failing with error:
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

Homebrew-core recently upgraded to cmake 4.0, which requires cmake statements for cmake>3.5


Crola1702:
Testing on: [Build Status](https://build.osrfoundation.org/job/gz_tools-ci-pr_any-homebrew-amd64/45/)

scpeters (approved):

PR #1308, opened 2025-04-21
Remove qt-main and qwt dependencies in noble-like conda env

iche033:
We finished migration to Qt6 in the source repos so removing qt-main (Qt5) dependencies.

Done by `pixi rm qt-main`. I also had to regenerate the lock file for the qt-main deps to be removed in that file.


j-rivero:
Oh I see, the `remove` command does not update the lock file. There should be no big problem on updating the dependencies or otherwise we need to pin the versions there. There could be regressions but I think that we are in a good moment for Jetty to update the list here.

traversaro:
> Oh I see, the `remove` command does not update the lock file. There should be no big problem on updating the dependencies or otherwise we need to pin the versions there. There could be regressions but I think that we are in a good moment for Jetty to update the list here.

If I recall well, in general any command that changes the pixi.lock file minimize the difference w.r.t. to the existing lock file. So if you want to create a lock file as you only had the pixi.toml file, it is better to before remove the existing pixi.lock . @wolfv @ruben-arts feel free to correct me if I got something wrong!

iche033:
Got it thanks for the feedback and info about how the remove command works! For this PR, I'll limit it to just removing the dependencies we don't need, and that was done successfully following @j-rivero's instructions to remove `qwt` along with `qt-main`. We don't need `qwt` anymore. We can generate a new lock file with updated dependencies and test that in a separate PR when we are ready for it.

j-rivero:
> > 
> 
> If I recall well, in general any command that changes the pixi.lock file minimize the difference w.r.t. to the existing lock file. So if you want to create a lock file as you only had the pixi.toml file, it is better to before remove the existing pixi.lock . @wolfv @ruben-arts feel free to correct me if I got something wrong!

Agree with anything in here. We tried to minimize the updates to the lock file when doing these cherry-pick additions / removals and `pixi add` and `pixi remove` has been working fine for that.

j-rivero (approved):

PR #1323, opened 2025-05-22
Crola1702/24.04 migration

Crola1702:
# Description

> [!WARNING]
> This shouldn't be merged until the jenkins server has been completely migrated to 24.04

This PR updates Jenkins DSL configuration to be able to run in the latest version of DSL plugin. It mostly migrates from unexsistent `postBuildScripts` syntax to `postBuildScript`

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFLinuxBuildPkgBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFLinuxBuildPkgBase.groovy
@@ -28,21 +28,26 @@ class OSRFLinuxBuildPkgBase
          }
       }
 
-      publishers
-      {
-        postBuildScripts {
-          steps {
-            shell("""\
-              #!/bin/bash -xe
+      publishers {
+        postBuildScript {
+          markBuildUnstable(false)
+          buildSteps {
+            postBuildStep {
+              results(['SUCCESS', 'NOT_BUILT', 'ABORTED', 'FAILURE', 'UNSTABLE'])
+              role('BOTH')
+              stopOnFailure(false)
+              buildSteps {
+                shell {
+                  command("""\
+                    #!/bin/bash -xe
 
-              [[ -d \${WORKSPACE}/pkgs ]] && sudo chown -R jenkins \${WORKSPACE}/pkgs
-              """.stripIndent())
+                    [[ -d \${WORKSPACE}/pkgs ]] && sudo chown -R jenkins \${WORKSPACE}/pkgs
+                    """.stripIndent())
+                }
+              }
+            }
           }
-
-          onlyIfBuildSucceeds(false)
-          onlyIfBuildFails(false)
         }
-
         archiveArtifacts('pkgs/*')
       }
     }
--- a/jenkins-scripts/dsl/_configs_/OSRFSourceCreation.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFSourceCreation.groovy
@@ -140,33 +140,62 @@ class OSRFSourceCreation
     job.with
     {
       publishers {
-        postBuildScripts {
-          steps {
-            conditionalSteps {
-             condition {
-              not {
-                expression('none|None|^$','${ENV,var="UPLOAD_TO_REPO"}')
+        postBuildScript {
+          buildSteps {
+            postBuildStep {
+              result(['SUCCESS'])
+              role('BOTH')
+              conditionalBuilder {
+                runner {
+                  fail()
                 }
-              }
-              steps {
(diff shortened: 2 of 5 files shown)


j-rivero:
> This is great ! One of the things that we can start doing right now is to move DSL changes to production so the PR is reduced and we are testing small pieces during the way.

Dah my bad sorry. Dynamic DSL that is not found in the API viewer. Need to check if that can work with the jar files :(

j-rivero:
@Crola1702 I've fixed all the conflicts. The generation is now failing because we are using DynamicDSL in the nex OSRFDslBase and generate_dsls files instead of native DSL. I've changed OSRFDslBase in [7563e40](https://github.com/gazebo-tooling/release-tools/pull/1323/commits/7563e40915a8e73e01eb44b1f4c1e71a80db4402) but please finish the migration so the CI can run. You have the API in https://jenkinsci.github.io/job-dsl-plugin and you can use the local generator in this repository inside the dsl/tools/ directory.

Crola1702:
> Be sure that deploying this branch is good on the testing server with respect to how the XML is generated by the configure blocks and how the real XML is in the current jobs.

So I got the diff of all jobs:

- [jenkins_jobs_diff.txt](https://github.com/user-attachments/files/21060524/jenkins_jobs_diff.txt)

Here is an enhanced version of the summary ChatGPT gave about the most important changes

Category | Summary
-- | --
Branch update (we should leave master branch instead) | All jobs now use a migration-specific branch
DSL plugin upgrades (some plugin pins and explicit fields with new dsl) | Better configuration and error handling for DSL jobs
SCM performance tweaks (adding clone options) | Faster and more reliable Git operations
Post-build behavior (Added `<stopOnFailure>false</stopOnFailure>` and moved conditions to `role="BOTH"` instead of only on `FAILURE`)  |  More robust handling of conditions and job outcomes
Node label renaming (use `built-in` instead of `master`) |  Reflects modern Jenkins naming conventions (built-in instead of master)
Email & retry plugin versions | Improves notification and retry behavior
Job Rotation and Retention Policies | Now consistently applied across many jobs


Just for the record:
<details><summary>How I got the full diff</summary>

1. Get remote config jobs: 
   ```bash
   # jenkins instance 
   mkdir -p /home/ubuntu/remote_configs/
   cd /var/lib/jenkins/jobs/
   find . -maxdepth 2 -name "config.xml" -exec sh -c 'cp {} /home/ubuntu/remote_configs/$(dirname {}).xml' \;
   chown -R ubuntu:ubuntu /home/ubuntu/remote_configs/` And then 

   # local machine
   scp -r ubuntu@citest.build.osrfoundation.org:/home/ubuntu/remote_configs .
   ```
2. Generate local dsls `java -jar tools/jobdsl.jar *.dsl`
3. `./compare_jobs_diff.sh`
    ```bash
    #!/usr/bin/env bash

    local_jobs_folder=${1:-jobs_config}
    remote_jobs_folder=${2:-remote_configs}

    mkdir -p diffs_folder
    rm -f jenkins_jobs_diff.txt
    touch jenkins_jobs_diff.txt

    for job_config in $local_jobs_folder/*.xml; do
      job_name=$(basename $job_config .xml)
      echo "Processing $job_name"
      if [ -f "$remote_jobs_folder/$job_name.xml" ]; then
        local_normalized=$(mktemp)
        remote_normalized=$(mktemp)

        xmllint --format --noblanks "$local_jobs_folder/$job_name.xml" | sed '1d' > "$local_normalized" 2>/dev/null
        xmllint --format --noblanks "$remote_jobs_folder/$job_name.xml" | sed '1d' > "$remote_normalized" 2>/dev/null

        echo "$job_name.xml" >> jenkins_jobs_diff.xml
        job_diff=$(diff -u "$local_normalized" "$remote_normalized")
        echo "$job_diff" > "diffs_folder/$job_name.diff"
        echo "$job_diff" >> jenkins_jobs_diff.xml
      else
        echo "$remote_jobs_folder/$job_name.xml not exists"
      fi
    done
    ```

</details>

j-rivero:
> Generate local dsls `java -jar tools/jobdsl.jar *.dsl`

There is an error here. What we are trying to know is if the XML that we injected using the configure block is read correctly by the real Jenkins Plugins on the server: we generate custom XML using the configure block but the local XML generator has no idea if that XML can be read by Jenkins or not.

The check to perform is: save the previous XML, run this branch in the generation of DSL code inside Jenkins, reload Jenkins and compare the XML generated by the real Jenkins and see if there are problems on the job /configure with respect to the values we are injected in the configure blocks.

j-rivero:
I've the patch ready, will merge it tomorrow.

PR #1322, opened 2025-05-22
[WIP] 24.04 migration development

Crola1702:

PR #1327, opened 2025-06-05
DSL: change PostBuildSCript from native to configure block

j-rivero:
The PR replaces the use of native postBuild DSL to use the configure block instead. This is related to #1322 and it is caused because the postBulild plugin does not support native DSL anymore starting with version 3.0.0.

We lost a bit of readability but we prevent the whole lost of all the local XML configuration, the CI implementation and a good bunch of QA tests which is a no go.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFLinuxBuildPkgBase.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFLinuxBuildPkgBase.groovy
@@ -30,17 +30,21 @@ class OSRFLinuxBuildPkgBase
 
       publishers
       {
-        postBuildScripts {
-          steps {
-            shell("""\
-              #!/bin/bash -xe
-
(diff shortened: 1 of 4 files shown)


scpeters:
I think this has been causing nightly debbuilds to fail

https://build.osrfoundation.org/job/gz-cmake5-debbuilder/99/

~~~
[PostBuildScript] - [INFO] Executing post build scripts.
[gz-cmake5-debbuilder] $ /bin/sh -xe /tmp/jenkins13657845013251628584.sh
+ [[ -d /home/jenkins/workspace/gz-cmake5-debbuilder/pkgs ]]
/tmp/jenkins13657845013251628584.sh: 3: [[: not found
~~~

scpeters:
> ```
> [gz-cmake5-debbuilder] $ /bin/sh -xe /tmp/jenkins13657845013251628584.sh
> + [[ -d /home/jenkins/workspace/gz-cmake5-debbuilder/pkgs ]]
> /tmp/jenkins13657845013251628584.sh: 3: [[: not found
> ```

it is running the script with `sh` but `[[` is a bash feature

PR #1328, opened 2025-06-09
OSRFLinuxBuildPkg dsl: use sh syntax

scpeters:
Try to fix debbuilds (follow up to https://github.com/gazebo-tooling/release-tools/pull/1327#issuecomment-2954284977) by replacing `[[` / `]]` with `[` / `]` for compatibility with `sh`.


j-rivero:
Ouch, thanks for the fix !

PR #1330, opened 2025-06-09
[Crola1702/24.04] Merge master changes and new local generator for DSL

j-rivero:
The PR should bring the changes in #1327 from master and produce a new updated copy of the local generator for CI and local development.


Crola1702 (changes requested):
Ahh I think CI is complaining because `cloneOption` is dynamic DSL, so we should use the `configure` block, or use the suggested `cloneOptions` instead (see [dsl api](https://build.osrfoundation.org/plugin/job-dsl/api-viewer/index.html#path/job-scm-git-extensions-cloneOption))

j-rivero:
Going to merge this branch @Crola1702 . You will need to check run the DSL again from your branch in citest to see if everything is fine there.

PR #1332, opened 2025-06-11
Parametrize `PACKAGES_URL` in multiple scripts

Crola1702:
As the title says


j-rivero (approved):
Nice. I usually prefer VAR:-value instead of VAR-value just to handle the cases of empty variables ended up in random wrong command calls. Feel free to change and merge.

PR #1338, opened 2025-07-04
Copy generate dsls from #1323

Crola1702:
This PR creates `generate_dsls.dsl` which is a generator of the `_dsl_*` jobs that configure other jobs

### DIFF
--- a/jenkins-scripts/dsl/generate_dsls.dsl
+++ b/jenkins-scripts/dsl/generate_dsls.dsl
@@ -35,7 +35,7 @@ dsl_triggerer.with {
   }
 
   triggers {
-    # run it at a fixed time in the night when the nightlies are build
+    // Run it at a fixed time in the night when the nightlies are build
     scm(Globals.CRON_START_NIGHTLY) {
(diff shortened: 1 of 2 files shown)


jenkins-scripts/dsl/generate_dsls.dsl
```diff
@@ -0,0 +1,57 @@
...
+dsl_triggerer.with {
+  label(Globals.nontest_label("built-in"))
+
+  parameters {
+    stringParam('RUN_DESCRIPTION', 'Automatic update',
+                'Description about why the run started')
+  }
+
+  triggers {
+    scm('@daily') {
```
j-rivero, line 38:
```suggestion
    # run it at a fixed time in the night when the nightlies are build
    scm(Globals.CRON_START_NIGHTLY) {
```

j-rivero: Instead of running daily with an uncontrolled hour let fix that to run every night so if things fail we know that it is coming from night changes not during day activities.

j-rivero (approved):

PR #1353, opened 2025-07-17
OSRFSourceCreation: calculate tarball sha256

scpeters:
Targeted at #1352, part of passing sha256 values to the debbuild and homebrew_pull_request_updater jobs.


j-rivero (changes requested):
Still failing, we need to fix the tarball path https://build.osrfoundation.org/job/_test_gz_source/98/console

PR #1356, opened 2025-07-21
Complete sha256 support in release: _releasepy job and -debbuilders

j-rivero:
The PR complete the sha256 support after #1351 and #1353 .

Two commits to implement:
 * _releasepy job support: not optional since it currently only handle the calls using `--source-tarball-uri` (not nightlies) so all the upstream callers can provide sha256.
 * The other support is for the `OSRFLinuxBuildPkg.groovy` that creates the -debbuilders. In this case it is optional since it needs to deal nightlies that do not provide a `SOURCE_TARBALL_URI` but builds the sources directly in the same build.


j-rivero:
Tested:

 * gz_source https://build.osrfoundation.org/job/_test_gz_source/104/
 * test_relese_py https://build.osrfoundation.org/job/_test_releasepy/63/console (displays the sha param in release.py calls)

We can also add the checksum to the repository_uploader although the full chain will be already covered by the builders in linux and brew which I think it is the most useful act.

scpeters (approved):

PR #1359, opened 2025-07-28
Add YATM configuration files for creating issues for managing collection releases

azeey:
We are going to try using YATM to maintain the list of tasks for a new Gazebo collection release. We used to use meta tickets such as https://github.com/gazebo-tooling/release-tools/issues/1092, but that has proved to be difficult to manage since there's a lot of tasks and it is difficult to divide up tasks inside one ticket. Going forward, the plan is to use YATM to create issues on the meta package (e.g. gazebosim/gz-ionic) and a Github Project to track the release process. This PR adds the YATM configuration and requirements file that contains the issues/tasks.

The first release we'll be using this on will be [gz-jetty](https://github.com/gazebosim/gz-jetty/issues)

### DIFF
--- a/jenkins-scripts/docker/lib/boilerplate_prepare.sh
+++ b/jenkins-scripts/docker/lib/boilerplate_prepare.sh
@@ -156,6 +156,12 @@ if ${NEED_SQUID_WORKAROUND}; then
   sudo service squid-deb-proxy restart
 fi
 
+# Required for asan to work on Noble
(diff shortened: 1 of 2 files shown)


scpeters:
with this change, I think we would be deprecating the following issue template:

* https://github.com/gazebo-tooling/release-tools/blob/master/.github/ISSUE_TEMPLATE/release_collection.md

if so, should we add a deprecation message to its title or delete it?

j-rivero (approved):
ready to go

PR #1388, opened 2025-09-08
Update pixi.toml to use qt6 6.8

azeey:
Part of https://github.com/gazebosim/gz-sim/issues/2945

### DIFF
--- a/conda/envs/noble_like/pixi.toml
+++ b/conda/envs/noble_like/pixi.toml
@@ -8,18 +8,16 @@ platforms = ["linux-64", "linux-aarch64", "win-64"]
 
 [dependencies]
 assimp = "*"
-bullet-cpp = "3.25.*"  # compatible with dart, good for noble
+bullet-cpp = "3.25.*"        # compatible with dart, good for noble
 cli11 = ">=2.4.1,<3"
 cppzmq = "*"
 curl = "*"
-dartsim = "6.13.2.*" # good for noble
-eigen = "3.4.0.*"  # good for noble
+dartsim = "6.13.*"           # good for noble
+eigen = "3.4.0.*"            # good for noble
 ffmpeg = "*"
 freeimage = "*"
-gdal = "3.8.4.*"
+gdal = "3.*"
 gflags = "*"
-glib = "*"
-gts = "*"
 jsoncpp = "*"
 libprotobuf = "*"
 libsqlite = "*"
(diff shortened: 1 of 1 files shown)


j-rivero:
> I am a proponent of having a pixi.toml file per release. It requires more changes in the infra, but once set up, we wouldn't need to update the toml files too often. And if we do, which would typically be at the beginning of a new release, we wouldn't have to worry about other releases.

Reviewing the situation our `win_conda_noble` is being used in Jetty and __upcoming__ so previous releases should not be affected by this qt change particularly.

Thinking more broadly, yes, we can probably maintain one file per Gazebo release, named them with the same name than the release so instead of using the directory name `noble_like` it will become `jetty` inside `conda/envs`. The only trade off is that we lost a bit of naming context (no reference to Ubuntu Noble version) and the we will need to handle the `_upcoming_` that most of the times probably want to use the latest release.

I drafted the changes https://github.com/gazebo-tooling/release-tools/pull/1393

conda/envs/noble_like/pixi.toml
```diff
@@ -6,20 +6,22 @@ authors = ["Jose Luis Rivero <jrivero@openrobotics.org>"]
 channels = ["conda-forge"]
 platforms = ["linux-64", "linux-aarch64", "win-64"]
 
+[target.win-64.activation.env]
+QT_QPA_PLATFORM_PLUGIN_PATH="%cd%\\.pixi\\envs\\default\\Library\\lib\\qt6\\plugins\\platforms"
+QML2_IMPORT_PATH="%cd%\\.pixi\\envs\\default\\Library\\lib\\qt6\\qml"
```
azeey, line 11: nit: Can we use `CONDA_PREFIX` here?

https://pixi.sh/dev/reference/environment_variables/#environment-variables-set-by-pixi

conda/envs/noble_like/pixi.toml
```diff
@@ -6,20 +6,22 @@ authors = ["Jose Luis Rivero <jrivero@openrobotics.org>"]
 channels = ["conda-forge"]
 platforms = ["linux-64", "linux-aarch64", "win-64"]
 
+[target.win-64.activation.env]
+QT_QPA_PLATFORM_PLUGIN_PATH="%cd%\\.pixi\\envs\\default\\Library\\lib\\qt6\\plugins\\platforms"
```
azeey, line 10: Nice! I like this approach while we wait for an upstream fix.

PR #1390, opened 2025-09-14
Add job to remove homebrew bottles for specified formulae and dependents

scpeters:
Edit: some changes were split out into https://github.com/gazebo-tooling/release-tools/pull/1401, and this is retargeted at that branch. Please review https://github.com/gazebo-tooling/release-tools/pull/1401 first.

Alternative to https://github.com/osrf/homebrew-simulation/issues/2606 that is implemented as a Jenkins job rather than a GitHub actions workflow.

This creates a Jenkins job that opens a pull request to the https://github.com/osrfbuild/homebrew-simulation fork to remove broken bottles for specified broken formulae and dependents.

This changes the required input variables to the `_homebrew_github_commit.bash` script so that `PULL_REQUEST_TITLE` is specified directly instead of indirectly using the `PACKAGE_ALIAS` and `VERSION` variables. Also, a `SKIP_COMMIT` variable is checked before trying to commit since `brew bump-revision` creates commits already with decent commit messages.

### Testing

* [Build Status](https://build.osrfoundation.org/job/_dsl_brew_release/1022/) https://build.osrfoundation.org/job/_dsl_brew_release/1022/
* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_remove_dependent_bottles/1/) https://build.osrfoundation.org/job/generic-release-homebrew_remove_dependent_bottles/1/
* Example PR: https://github.com/osrf/homebrew-simulation/pull/3221

PR #1419, opened 2025-12-04
brew: fix SKIP_COMMIT logic

scpeters:
I introduced a bug in https://github.com/gazebo-tooling/release-tools/pull/1390, since the pull_request_updater and bottle_hash_updater jobs don't set `SKIP_COMMIT` to anything and the new logic prevents those jobs from committing. This updates the logic to always set `SKIP_COMMIT` to true or false in brew pull request scripts.

### Test showing failure with master branch

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1742/) https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1742/

### Test showing success with this branch

* [Build Status](https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1743/) https://build.osrfoundation.org/job/generic-release-homebrew_pull_request_updater/1743/

### DIFF
--- a/jenkins-scripts/lib/_homebrew_github_commit.bash
+++ b/jenkins-scripts/lib/_homebrew_github_commit.bash
@@ -8,9 +8,13 @@
 PR_URL_export_file=${PR_URL_export_file:-${WORKSPACE}/pull_request_created.properties}
 
 echo '# BEGIN SECTION: check variables'
-if [ -z "${COMMIT_MESSAGE}" ]; then
-  if [ -z "${SKIP_COMMIT}" ]; then
-    echo One of COMMIT_MESSAGE or SKIP_COMMIT must be specified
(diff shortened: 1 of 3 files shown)


scpeters:
> Sounds a bit weird to me that we are sourcing a file called `jenkins-scripts/lib/_homebrew_github_commit.bash` for not doing a commit. Is that the file needs to have a new name or we need to source it conditionally in a different place?

the name is not completely accurate. the only thing it does for sure is [git push](https://github.com/gazebo-tooling/release-tools/blob/4f88ff023197ae2c8fda45a4601a02a950a9feed/jenkins-scripts/lib/_homebrew_github_commit.bash#L63). It can also call `git commit` and/or open a github pull request, but it doesn't always do that

Issue #1425, opened 2025-12-18
New collection: Kura


azeey:
- [ ] Create `gazebodistro` file for new collection ([example](https://github.com/gazebo-tooling/gazebodistro/pull/181))
  - https://github.com/gazebo-tooling/gazebodistro/pull/225
- [ ] Create `gz-<collection>` and `gz-<collection>-release` repositories ([example source](https://github.com/gazebosim/gz-jetty), [example release](https://github.com/gazebo-release/gz-jetty-release))
- [ ] Create homebrew-simulation formula for new collection ([example](https://github.com/osrf/homebrew-simulation/pull/2832))
- [ ] Create collection view on release-tools and build nightlies ([example](https://github.com/gazebo-tooling/release-tools/pull/1199))
- [ ] Add docs for new collection ([example](https://github.com/gazebosim/docs/pull/524))
- [ ] Add new collection label:
    - [ ] Add to pr-collection-labeler ([example](https://github.com/gazebo-tooling/pr-collection-labeler/pull/36))
    - [ ] Add to all libraries
      Example script:`cat collection-kura.yaml | sed -n "s#.*https://github.com/\(.*\)#\1#p" | xargs -L1 gh label create "🏯 kura" -R`

PR #1427, opened 2025-12-18
Add kura collection to gz-collections

azeey:
Part of https://github.com/gazebo-tooling/release-tools/issues/1425

This simply copies the Jetty configuration for now.

Replaces #1426 becasue CI doesn't support PRs from forks.


jenkins-scripts/dsl/gz-collections.yaml
```diff
@@ -365,6 +365,90 @@ collections:
...
+          current_branch: gz-sensors10
+      - name: gz-physics
+        major_version: 9
+        repo:
+          current_branch: gz-physics9
+      - name: gz-sim
+        major_version: 10
+        repo:
+          current_branch: gz-sim10
+      - name: gz-launch
```
iche033: note that we can exclude gz-launch now or we can do it when bumping major versions.

iche033 (approved):

PR #1467, opened 2026-03-24
Remove large-memory requirement

j-rivero:
This goes together with the change on standard agents in Windows and Linux to 16Gb.

This pull request removes the use of the `large-memory` label from Jenkins job configurations and documentation, consolidating requirements to use only the `docker` label for relevant jobs. It also updates documentation to specify memory requirements directly in the description for clarity. These changes simplify label management and make node requirements more explicit.

**Jenkins job label and requirements simplification:**

* Removed the `large_memory` requirement for `gz-physics` from all relevant sections in the `gz-collections.yaml` configuration.
* Removed the `large-memory` label from all Jenkins job definitions in DSL scripts (`debian.dsl`, `extra.dsl`, `gzdev.dsl`), so jobs now only require the `docker` label. 
* Updated the label assignment backstop script to remove `large-memory` from nightly label definitions, leaving only the `docker` label for the noble release.

**Documentation updates:**

* Updated `BUILDFARM_NODES.md` to specify that both `docker` and `win` nodes require 16GB RAM, moving the memory requirement into the description instead of using a separate `large-memory` label.

**Script and logic adjustments:**

* Removed checks and code paths related to the `large-memory` label from the DSL check script and the release script, since the label is no longer in use.

### DIFF
--- a/release.py
+++ b/release.py
@@ -920,14 +920,10 @@ def go(argv):
                         # Need to use JENKINS_NODE_TAG parameter for large memory nodes
                         # since it runs qemu emulation
                         linux_platform_params['JENKINS_NODE_TAG'] = 'linux-' + a
-                    elif ('ignition-physics' in args.package_alias) or \
-                         ('gz-physics' in args.package_alias):
-                        linux_platform_params['JENKINS_NODE_TAG'] = 'large-memory'
 
                     # control nightly generation using a single machine to process
                     # all distribution builds to avoid race conditions. Note: this
-                    # assumes that large-memory nodes are being used for nightly
-                    # tags.
+                    # assumes that nodes are being tagged for nightly tags.
                     # https://github.com/gazebo-tooling/release-tools/issues/644
                     if (NIGHTLY):
                         assert a == 'amd64', f'Nightly tag assumed amd64 but arch is {a}'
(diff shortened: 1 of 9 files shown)


release.py
```diff
@@ -920,14 +920,10 @@ def go(argv):
...
                         linux_platform_params['JENKINS_NODE_TAG'] = 'linux-' + a
-                    elif ('ignition-physics' in args.package_alias) or \
-                         ('gz-physics' in args.package_alias):
-                        linux_platform_params['JENKINS_NODE_TAG'] = 'large-memory'
 
                     # control nightly generation using a single machine to process
                     # all distribution builds to avoid race conditions. Note: this
-                    # assumes that large-memory nodes are being used for nightly
-                    # tags.
+                    # assumes that nodes are being tagged for nightly tags.
```
azeey: nit: Consider removing the sentence starting "Note: ". As it's written now, I'm quite sure what it means.

azeey (approved):
LGTM!

release.py
j-rivero: re-write it here [5936c57](https://github.com/gazebo-tooling/release-tools/pull/1467/commits/5936c57570ebb1e89e7168ef52f1d24363fd77c5)

PR #1470, opened 2026-03-26
fix: Local build to change cwd to temp workspace before running colcon

azeey:
This also removes desktop notification from colcon commands, which are unnecessary in CI environments.


j-rivero:
Tested [Build Status](https://build.osrfoundation.org/job/gz_common-pr-cnlwin/203/)

Issue #1442, opened 2026-02-12
:farmer: gz_sim_{10,main}_cnlwin gz_physics-ign-physics5-clwin fails investigation


miguelgonrod:
### Reference Builds
- https://build.osrfoundation.org/job/gz_physics-ign-physics5-clwin/100/
- https://build.osrfoundation.org/job/gz_sim-10-cnlwin/52
- https://build.osrfoundation.org/job/gz_sim-10-cnlwin/53
- https://build.osrfoundation.org/job/gz_sim-10-cnlwin/54

### gz_sim-main
Comparing [#101](https://build.osrfoundation.org/job/gz_sim-main-cnlwin/101/) (passing) and [#102](https://build.osrfoundation.org/job/gz_sim-main-cnlwin/102/) (Failing)

<details><summary>Check version differences log</summary>

```diff
- gz-physics version 9.0.0
+ gz-physics version 9.1.0

- Looking for gz-rendering -- found version 10.0.0
+ Looking for gz-rendering -- found version 10.0.1

- Looking for gz-physics -- found version 9.0.0
+ Looking for gz-physics -- found version 9.1.0

- Looking for gz-physics-heightmap -- found version 9.0.0
+ Looking for gz-physics-heightmap -- found version 9.1.0

- Looking for gz-physics-mesh -- found version 9.0.0
+ Looking for gz-physics-mesh -- found version 9.1.0

- Looking for gz-physics-sdf -- found version 9.0.0
+ Looking for gz-physics-sdf -- found version 9.1.0

- Looking for gz-sensors -- found version 10.0.0
+ Looking for gz-sensors -- found version 10.0.1

- Looking for gz-rendering -- found version 10.0.0
+ Looking for gz-rendering -- found version 10.0.1

- Looking for gz-sensors-air_pressure -- found version 10.0.0
+ Looking for gz-sensors-air_pressure -- found version 10.0.1

- Looking for gz-sensors-air_speed -- found version 10.0.0
+ Looking for gz-sensors-air_speed -- found version 10.0.1

- Looking for gz-sensors-altimeter -- found version 10.0.0
+ Looking for gz-sensors-altimeter -- found version 10.0.1

- Looking for gz-sensors-imu -- found version 10.0.0
+ Looking for gz-sensors-imu -- found version 10.0.1

- Looking for gz-sensors-force_torque -- found version 10.0.0
+ Looking for gz-sensors-force_torque -- found version 10.0.1

[...]

PR #1477, opened 2026-04-08
Move pixi paths to %PROGRAMDATA% to prevent cleanup issues

j-rivero:
A possible fix for https://github.com/gazebo-tooling/release-tools/issues/1442

As described in  , the main hypothesis is that "Storage Sense" and/or "SilentCleanup" Windows mechanisms could be trimming files from C:\Windows\Temp\pixi\project\.pixi\envs\default\ during long builds so the pull request updates the handling of temporary and project paths for the Pixi environment in both the Windows batch script and the Python build script. The main change is to move Pixi-related directories from the user-specific `%TMP%` directory to the system-wide `%PROGRAMDATA%` directory, ensuring greater consistency and potentially avoiding issues with user-specific temp directories.

**Environment variable and path changes:**

* Updated `PIXI_PROJECT_PATH`, `PIXI_BOOTSTRAP_PROJECT_PATH`, and `PIXI_TMPDIR` in `windows_env_vars.bat` to use `%PROGRAMDATA%\pixi` instead of `%TMP%\pixi` for system-wide consistency.
* Modified the `local_build.py` script to set `PIXI_PROJECT_PATH` using the `PROGRAMDATA` environment variable instead of `TMP`, aligning with the batch script changes.

Testing run: [Build Status](https://build.osrfoundation.org/job/gz_common-pr-cnlwin/210/)

### DIFF
--- a/jenkins-scripts/lib/windows_env_vars.bat
+++ b/jenkins-scripts/lib/windows_env_vars.bat
@@ -2,11 +2,13 @@
 set "PIXI_VERSION=0.44.0"
 set "PIXI_URL=https://github.com/prefix-dev/pixi/releases/download/v%PIXI_VERSION%/pixi-x86_64-pc-windows-msvc.exe"
 if NOT DEFINED PIXI_PROJECT_PATH (
-  set "PIXI_PROJECT_PATH=%TMP%\pixi\project"
+  set "PIXI_PROJECT_PATH=%LOCALAPPDATA%\pixi\project"
 )
-set "PIXI_BOOTSTRAP_PROJECT_PATH=%TMP%\pixi\bootstrap_project"
+if NOT DEFINED PIXI_BOOTSTRAP_PROJECT_PATH (
+  set "PIXI_BOOTSTRAP_PROJECT_PATH=%LOCALAPPDATA%\pixi\bootstrap_project"
+)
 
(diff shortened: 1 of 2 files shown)


azeey (approved):
LGTM!

Unrelated: Would it be possible to include the installation of `pixi` in the node provisioning scripts so that it doesn't have to be installed in every job run?

j-rivero:
> Unrelated: Would it be possible to include the installation of `pixi` in the node provisioning scripts so that it doesn't have to be installed in every job run?

Yep. There are pros/cons on that decission: we gain time in the CI builds, we lost the flexibility of upgrading the pixi versions quickly by doing changes to this repository. We could find an intermediate solution.

azeey:
> > Unrelated: Would it be possible to include the installation of `pixi` in the node provisioning scripts so that it doesn't have to be installed in every job run?
> 
> Yep. There are pros/cons on that decission: we gain time in the CI builds, we lost the flexibility of upgrading the pixi versions quickly by doing changes to this repository. We could find an intermediate solution.

It is true we lose the flexibility, but considering we're losing time on every CI build and we rarely need to upgrade pixi, I think it is worth it. Also, we can always do a `pixi upgrade` in this repo for testing before upgrading it in the provisioning scripts.

PR #1484, opened 2026-04-21
Use brew main branch until unbottled is fixed

scpeters:
The `brew unbottled` command is broken in 5.1.7 and was recently fixed in https://github.com/Homebrew/brew/pull/22050, so use `main` for a few weeks until the fix is released.

Needed since https://github.com/osrf/homebrew-simulation/pull/3416 was not properly generated

Properly generated https://github.com/osrf/homebrew-simulation/pull/3417 with this branch

### DIFF
--- a/jenkins-scripts/lib/_homebrew_github_setup.bash
+++ b/jenkins-scripts/lib/_homebrew_github_setup.bash
@@ -45,8 +45,8 @@ else
 fi
 # Use stable version of brew
 export HOMEBREW_UPDATE_TO_TAG=1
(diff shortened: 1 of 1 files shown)


scpeters:
> The `brew unbottled` command is broken in 5.1.7 and was recently fixed in [Homebrew/brew#22050](https://github.com/Homebrew/brew/pull/22050), so use `main` for a few weeks until the fix is released.

fix was released in [brew 5.1.8](https://github.com/Homebrew/brew/releases/tag/5.1.8)

Issue #1485, opened 2026-04-22
Enable CI on 26.04 for Rotary, Jetty


scpeters:
Ubuntu 26.04 Resolute will be released soon, so we should enable CI support for this distribution for Rotary and Jetty.

### Create 26.04 debian metadata for Rotary packages

- [x] gz-rotary-cmake
    - [x] https://github.com/gazebo-release/gz-rotary-cmake-release/pull/1
    - [x] https://github.com/gazebo-release/gz-rotary-cmake-release/pull/2
- [x] gz-rotary-utils https://github.com/gazebo-release/gz-rotary-utils-release/pull/1
- [ ] gz-rotary-math
- [ ] gz-rotary-plugin
- [ ] gz-rotary-tools
- [ ] 

### Enable action-gz-ci on 26.04 for Rotary packages

* [x] action-gz-ci https://github.com/gazebo-tooling/action-gz-ci/pull/88
* [x] gz-rotary-cmake https://github.com/gazebosim/gz-cmake/pull/549
* [x] gz-rotary-utils https://github.com/gazebosim/gz-utils/pull/208
* [x] gz-rotary-math https://github.com/gazebosim/gz-math/pull/746
* [ ] gz-rotary-plugin
* [ ] gz-rotary-tools
* [ ]

scpeters:
first debbuild attempt of gz-rotary-cmake failed

* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz-rotary-cmake-debbuilder&build=47)](https://build.osrfoundation.org/job/gz-rotary-cmake-debbuilder/47/) https://build.osrfoundation.org/job/gz-rotary-cmake-debbuilder/47/

~~~
Step 53/58 : RUN newgrp docker
 ---> Running in 0d088d3c1538
[91m/bin/sh: 1: newgrp: not found
[0mThe command '/bin/sh -c newgrp docker' returned a non-zero code: 127
~~~

the `newgrp` utility is now [part of the util-linux-extra package on 26.04](https://packages.ubuntu.com/resolute/amd64/util-linux-extra/filelist), while it was [part of the login package on 24.04](https://packages.ubuntu.com/noble/amd64/login/filelist)

scpeters:
[...]

PR #1492, opened 2026-05-04
Script to copy metadata for new Ubuntu release

scpeters:
This adds a script for copying and renaming a debian metadata folder for every release repository in a given collection. Several pull requests have been opened using this script as part of #1485:

* https://github.com/gazebo-release/gz-rotary-math-release/pull/1
* https://github.com/gazebo-release/gz-rotary-common-release/pull/2
* etc

### DIFF
--- a/release-repo-scripts/collection_copy_ubuntu_metadata.bash
+++ b/release-repo-scripts/collection_copy_ubuntu_metadata.bash
@@ -1,0 +1,244 @@
+#!/usr/bin/env bash
+# Copyright (C) 2026 Open Source Robotics Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# The script will copy and rename an existing Ubuntu metadata folder in
+# the release repositories of a given collection to add support for a new
+# Ubuntu release.
+#
+# Requires the 'gh' CLI, `xmllint`, and 'python-vcstool' to be installed.
+#
+# Usage:
+# $ ./collection_copy_ubuntu_metadata.bash <collection> <existing_ubuntu> <new_ubuntu> <issue_reference>
+#
+# For example, to copy the noble folder to resolute in all Jetty release repositories:
+#
+# ./collection_copy_ubuntu_metadata.bash jetty noble resolute gazebo-tooling/release-tools/issues/1485
+#
+# Before committing to each repository, the script asks "Commit <repository name>?".
+# Before saying yes, navigate to the repository and check if the diff looks reasonable.
+# When you say yes, the changes will be committed and pushed. Click on the link printed
+# by GitHub to open the pull request.
+
+# default to false
+DRY_RUN=${DRY_RUN:-false}
+
+DEFAULT="\e[39m"
+DEFAULT_BG="\e[49m"
+
+GREY="\e[90m"
+GREEN="\e[32m"
+RED="\e[31m"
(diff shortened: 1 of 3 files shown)


scpeters:
one thing to add is to update the collection repo itself ([gz-rotary-release](https://github.com/gazebo-release/gz-rotary-release) for example)

scpeters:
> one thing to add is to update the collection repo itself ([gz-rotary-release](https://github.com/gazebo-release/gz-rotary-release) for example)

updated in https://github.com/gazebo-tooling/release-tools/pull/1492/commits/ba5a64150e7849203feda8691a9a5869ab3a913a and generated the following:

* https://github.com/gazebo-release/gz-rotary-release/pull/2

release-repo-scripts/collection_copy_ubuntu_metadata.bash
```diff
@@ -0,0 +1,249 @@
...
+  if ${DRY_RUN}; then
+    echo -e "${GREEN_BG}${REPO}: dry-run enabled (avoid commit and PR). Press to continue${DEFAULT_BG}"
+    read CONTINUE
+  else
+    echo -e "${GREEN_BG}${REPO}: Commit ${REPO} and open PR? (y/n)${DEFAULT_BG}"
+    read CONTINUE
+    if [ "$CONTINUE" = "y" ]; then
+      git commit -sam"${COMMIT_MSG_PREFIX}${COMMIT_MSG}"
+      git push origin ${CURRENT_BRANCH}
+      echo gh pr create --title "${COMMIT_MSG_PREFIX}${COMMIT_MSG}" --body "${PR_TEXT}" --repo ${ORG}/${REPO} --base ${BASE_BRANCH}
```
j-rivero:
```suggestion
```
Debugging code?

j-rivero (commented):
Script is mostly fine, the only point in my list is that it has a small duplication of the existing script `new_ubuntu_distribution.bash` https://github.com/gazebo-tooling/release-tools/blob/master/release-repo-scripts/new_ubuntu_distribution.bash when it creates new directories inside the -release repositories. The changes from one to other are not big:

```
  # new_ubuntu_distribution.bash
  cp -a "${latest_distro}" "${NEW_DISTRO}"
  dch --distribution="${NEW_DISTRO}" --force-distribution …
  sed -i -e "s:~${latest_distro}.*):~${NEW_DISTRO}):" debian/changelog
  head -n 5 debian/changelog > debian/changelog.new && mv …
  rm -f debian/format

  vs

  # collection_copy_ubuntu_metadata.bash
  mkdir -p $NEW_DISTRO
  cp -R $OLD_DISTRO/* $NEW_DISTRO/
  sed -i -e "s@$OLD_DISTRO@$NEW_DISTRO@" $NEW_DISTRO/debian/changelog
```

but worth considering reusing the existing one to reduce the amount of duplicate code in this repo. I can take care of the changes although we would need testing.

release-repo-scripts/collection_copy_ubuntu_metadata.bash
scpeters: yes, this was for debugging purposes, let's remove

scpeters (commented):
> Script is mostly fine, the only point in my list is that it has a small duplication of the existing script `new_ubuntu_distribution.bash` ...
> ...
> but worth considering reusing the existing one to reduce the amount of duplicate code in this repo. I can take care of the changes although we would need testing.

that's fine with me; I didn't understand exactly what the existing script does, so I made my own. Please update as you see fit either here or in a new PR

scpeters:
> > Script is mostly fine, the only point in my list is that it has a small duplication of the existing script `new_ubuntu_distribution.bash` ...
> > ...
> > but worth considering reusing the existing one to reduce the amount of duplicate code in this repo. I can take care of the changes although we would need testing.
> 
> that's fine with me; I didn't understand exactly what the existing script does, so I made my own. Please update as you see fit either here or in a new PR

I think we are ready to start building some 26.04 debs for Jetty; you could use that as an opportunity to test your updates to the script

j-rivero:
Added documentation for this script and I think that now is better to commit so we don't lose it in a PR. If we found the need of merging with existing https://github.com/gazebo-tooling/release-tools/blob/master/release-repo-scripts/new_ubuntu_distribution.bash we can do it in a new PR.

j-rivero (approved):

PR #1510, opened 2026-06-08
dsl: remove unused PULL_REQUEST_URL brew param

scpeters:
This parameter is added to all brew compilation jobs but is unused. It has not been used since https://github.com/gazebo-tooling/release-tools/pull/575, and we discussed removing the param in https://github.com/gazebo-tooling/release-tools/pull/575#discussion_r758599949. I was against removing it at the time, though I can't remember why since the `PULL_REQUEST_URL` parameter is [independently defined](https://github.com/gazebo-tooling/release-tools/blob/a0d35164439648791330664658ff1bbf0335441e/jenkins-scripts/dsl/brew_release.dsl#L270) in the other job where it is used. Steve today thinks it's safe to remove.

### DIFF
--- a/jenkins-scripts/dsl/_configs_/OSRFBrewCompilation.groovy
+++ b/jenkins-scripts/dsl/_configs_/OSRFBrewCompilation.groovy
@@ -20,12 +20,6 @@ class OSRFBrewCompilation extends OSRFOsXBase
 
     job.with
     {
-      parameters
-      {
-        stringParam("PULL_REQUEST_URL", '',
-                    'Pull request URL (osrf/homebrew-simulation) pointing to a pull request.')
-      }
-
       publishers
       {
         configure { project ->


j-rivero:
I see the variable referenced from other places:

```
dsl/brew_release.dsl:              predefinedProp("PULL_REQUEST_URL", "https://github.com/osrf/homebrew-simulation/pull/\${ghprbPullId}")
dsl/brew_release.dsl:    stringParam("PULL_REQUEST_URL", '',
lib/_homebrew_github_commit.bash:# - PULL_REQUEST_TITLE or PULL_REQUEST_URL
lib/_homebrew_github_commit.bash:  if [ -z "${PULL_REQUEST_URL}" ]; then
lib/_homebrew_github_commit.bash:    echo One of PULL_REQUEST_TITLE or PULL_REQUEST_URL must be specified
lib/_homebrew_github_commit.bash:if [ -z "${PULL_REQUEST_URL}" ]; then
lib/_homebrew_github_commit.bash:  echo "PULL_REQUEST_URL=${PR_URL}" > ${PR_URL_export_file}
lib/_homebrew_github_commit.bash:  echo "Pull request updated: ${PULL_REQUEST_URL}"
lib/homebrew_bottle_pullrequest.bash:if [ -z "${PULL_REQUEST_URL}" ]; then
lib/homebrew_bottle_pullrequest.bash:  echo PULL_REQUEST_URL not specified
lib/homebrew_bottle_pullrequest.bash:PULL_REQUEST_API_URL=$(echo ${PULL_REQUEST_URL} \
```

The DSL is injected a predefinedProp when launching `bottle_hash_updater_job_name` but also adds a GUI paremeter. Is the intention to preserve these?

scpeters:
the `PULL_REQUEST_URL` parameter is defined in two disjoint sets of jobs:

* in jenkins-scripts/dsl/_configs_/OSRFBrewCompilation.groovy the parameter has been defined for all brew compilation jobs but has not been used since https://github.com/gazebo-tooling/release-tools/pull/575
* in jenkins-scripts/dsl/brew_release.dsl, the parameter is defined for use in bottle building / PR hash updating jobs and scripts. The parameter is still used here, so these scripts are unchanged

it's a little confusing because the same parameter name was used in two different sets of jobs, but the brew compilation jobs don't interact with the bottle building jobs in any way, so I believe it is safe to remove the parameter from `OSRFBrewCompilation.groovy` since it is no longer used

Issue #1410, opened 2025-11-12
Some homebrew CI builds get stuck


scpeters:
The following builds have gotten stuck after running for hours and need to be manually aborted:

* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz_gui-ci-gz-gui8-homebrew-arm64&build=16)](https://build.osrfoundation.org/job/gz_gui-ci-gz-gui8-homebrew-arm64/16/) https://build.osrfoundation.org/job/gz_gui-ci-gz-gui8-homebrew-arm64/16/
* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz_launch8-install_bottle-homebrew-arm64&build=59)](https://build.osrfoundation.org/job/gz_launch8-install_bottle-homebrew-arm64/59/) https://build.osrfoundation.org/job/gz_launch8-install_bottle-homebrew-arm64/59/
* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=gz_launch8-install_bottle-homebrew-arm64&build=61)](https://build.osrfoundation.org/job/gz_launch8-install_bottle-homebrew-arm64/61/) https://build.osrfoundation.org/job/gz_launch8-install_bottle-homebrew-arm64/61/

This has occurred with different jobs on different machines.

scpeters:
just noticed again in https://build.osrfoundation.org/job/gz_gui8-install_bottle-homebrew-arm64/254/console

~~~
11:28:58 ==> Fetching downloads for: gz-gui8
11:29:00 Warning: freeimage has been deprecated because it has outstanding CVEs and requires patches to build! It will be disabled on 2027-01-10.
11:29:09 Warning: qt@5 has been deprecated because it is not supported upstream! It will be disabled on 2027-05-19.
11:29:10 ✔︎ Bottle pkgconf (2.5.1)
...
11:29:12 ✔︎ Formula gz-gui8 (8.4.0)
11:29:13 ✔︎ Bottle aws-sdk-cpp (1.11.810)
14:52:46 Build was aborted
~~~

it was stuck for several hours, so I aborted the job

scpeters:
> just noticed again in https://build.osrfoundation.org/job/gz_gui8-install_bottle-homebrew-arm64/254/console
> 
> ```
> 11:28:58 ==> Fetching downloads for: gz-gui8
[...]

PR #1519, opened 2026-07-20
Add a default timeout for all jobs

Crola1702:
I've seen different kind of jobs hung in the buildfarm for a long time

This PR sets a default timeout for all jobs


j-rivero (commented):
Related to #1410 . It requires the installation of a new timeout plugin in Jenkins that is pending for the next upgrade.

PR #1521, opened 2026-07-23
arm: use official Ubuntu images instead of osrf custom ones

j-rivero:
The arm64 debbuilder was pinned to the manually-maintained osrf/ubuntu_arm64:${DISTRO} base images, built and pushed to Docker Hub by hand. The custom images were a 2016 workaround from before Docker offered native multi-arch base images and cross-arch emulation. armhf was already migrated off them (#647, #1135) using the official ubuntu:${DISTRO} image plus --platform.

Since Focal is EOL and no longer built, so its remaining osrf/ special case for armhf is dead code. Drop it and handle both arm arches with a single check.

Nowadays we are building images on native platforms so we can probably drop the platform argument for arm but I think that it does no hurt and it is robust if we decide to change to emulation at some point.

Assisted-By: Opus 4.8

### DIFF
--- a/jenkins-scripts/docker/lib/docker_generate_dockerfile.bash
+++ b/jenkins-scripts/docker/lib/docker_generate_dockerfile.bash
@@ -118,15 +118,8 @@ case ${ARCH} in
        FROM_VALUE=${ARCH}/${LINUX_DISTRO}:${DISTRO}
      fi
      ;;
-   'armhf')
-     if [[ ${DISTRO} == 'focal' ]]; then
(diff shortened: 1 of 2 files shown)


azeey (approved):
LGTM

scpeters:
I've queued a build of gz-cmake5 arm64 on resolute

[Build Status](https://build.osrfoundation.org/job/gz-cmake5-debbuilder/406/) https://build.osrfoundation.org/job/gz-cmake5-debbuilder/406/

Just need for weekend schedule to end and an arm build machine to be re-added

Issue #1503, opened 2026-05-22
Enable CI on 26.04 for Jetty


scpeters:
Support for Ubuntu 26.04 Resolute in Rotary is nearly complete (see https://github.com/gazebo-tooling/release-tools/issues/1485), so work on enabling CI support for Jetty as well.

### Create 26.04 debian metadata for Jetty packages

For reference, I open the collection_copy_metadata PRs with https://github.com/gazebo-tooling/release-tools/pull/1532

~~~
release-repo-scripts/collection_copy_ubuntu_metadata.bash jetty noble resolute gazebo-tooling/release-tools/issues/1503
~~~

- [x] Increase `cmake_minimum_required` version in debian tests (repeat what was done for rotary metadata in https://github.com/gazebo-tooling/release-tools/issues/1485)
    - [x]  https://github.com/gazebo-release/gz-cmake5-release/pull/6
    - [x]  https://github.com/gazebo-release/gz-fuel-tools11-release/pull/5
    - [x]  https://github.com/gazebo-release/gz-plugin4-release/pull/6
    - [x]  https://github.com/gazebo-release/gz-transport15-release/pull/10
    - [x]  https://github.com/gazebo-release/sdformat16-release/pull/10
- [x] cmake https://github.com/gazebo-release/gz-cmake5-release/pull/7
- [x] tools https://github.com/gazebo-release/gz-tools2-release/pull/24
- [x] utils https://github.com/gazebo-release/gz-utils4-release/pull/9
- [x] math https://github.com/gazebo-release/gz-math9-release/pull/8
- [x] common https://github.com/gazebo-release/gz-common7-release/pull/7
- [x] msgs https://github.com/gazebo-release/gz-msgs12-release/pull/6
- [x] plugin https://github.com/gazebo-release/gz-plugin4-release/pull/7
- [x] sdformat https://github.com/gazebo-release/sdformat16-release/pull/11
- [x] fuel-tools https://github.com/gazebo-release/gz-fuel-tools11-release/pull/6
- [x] rendering
  - [x] https://github.com/gazebo-forks/ogre-2.3-release/pull/11
  - [x] https://github.com/gazebo-release/gz-rendering10-release/pull/7
  - [x] https://github.com/gazebo-release/gz-rendering10-release/pull/6
- [x] transport https://github.com/gazebo-release/gz-transport15-release/pull/11
- [x] gui https://github.com/gazebo-release/gz-gui10-release/pull/9
[...]

PR #1532, opened 2026-08-09
Fix metadata copy script for Jetty

scpeters:
Some changes I've been making as part of #1503.

* Remove numbers from package name for gz-tools2
* Translate '_' to '-' for gz-fuel_tools
* Remove gz-${COLLECTION} before cloning to avoid cloning errors
* Replace changelog distro multiple times per line


scpeters:
I used these changes while copying resolute metadata from noble for jetty

j-rivero (approved):

